PR #8, opened 2020-09-16
[filtering] STALTA cleanups

jsaul:
This is a cleanup of the STALTA filters. Changes include:

* General cleanups and improved readability
* Improved behavior at the beginning of a segment, with the initialization time reduced to the length of the STA window. This has turned out to be enough and also robust.
* Addition of a non-recursive STALTA_Classic class.
* Updated math.i to generate the proper Python wrappers.

### DIFF
--- a/libs/seiscomp/math/filter/stalta.cpp
+++ b/libs/seiscomp/math/filter/stalta.cpp
@@ -17,58 +17,44 @@
  * gempa GmbH.                                                             *
  ***************************************************************************/
 
-
-#include<seiscomp/math/filter/stalta.h>
+#include <seiscomp/core/exceptions.h>
+#include <seiscomp/math/filter/stalta.h>
 
 namespace Seiscomp {
 namespace Math {
 namespace Filtering {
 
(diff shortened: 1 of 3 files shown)


libs/seiscomp/math/filter/stalta.cpp
```diff
@@ -169,11 +150,13 @@ template<typename TYPE>
...
+	if (fsamp <= 0.)
+		throw Core::ValueException("Sampling frequency must be positive");
 	_fsamp  = fsamp;
-	_numSTA = int(_lenSTA*fsamp+0.5);
-	_numLTA = int(_lenLTA*fsamp+0.5);
-	_initLength = _numLTA/2;
+	_numSTA = int(_lenSTA*_fsamp+0.5);
+	_numLTA = int(_lenLTA*_fsamp+0.5);
 	reset();
+std::cerr << "setSamplingFrequency fsamp=" << _fsamp << std::endl;
```
gempa-jabe, line 159: Please remove in final PR.

PR #10, opened 2020-10-29
Adding tooltips for costum columns in scolv

luca-s:
Dear Developers,

I would like to propose this change for the custom column option of scolv. When the custom column is created I would set the tooltip for that column to be the same as the content of the cell.

What do you think about it?

Regards,
Luca

### DIFF
--- a/libs/seiscomp/gui/datamodel/eventlistview.cpp
+++ b/libs/seiscomp/gui/datamodel/eventlistview.cpp
@@ -889,11 +889,13 @@ class OriginTreeItem : public SchemeTreeItem {
 
 			if ( config.customColumn != -1 ) {
 				setText(config.customColumn, config.customDefaultText);
+			  setToolTip(config.customColumn, config.customDefaultText);
 				setData(config.customColumn, Qt::ForegroundRole, QVariant());
 				if ( !config.originCommentID.empty() ) {
 					for ( size_t i = 0; i < ori->commentCount(); ++i ) {
 						if ( ori->comment(i)->id() == config.originCommentID ) {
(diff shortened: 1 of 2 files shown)


libs/seiscomp/gui/datamodel/eventlistview.cpp
```diff
@@ -889,11 +889,13 @@ class OriginTreeItem : public SchemeTreeItem {
 
 			if ( config.customColumn != -1 ) {
 				setText(config.customColumn, config.customDefaultText);
+			  setToolTip(config.customColumn, config.customDefaultText);
```
gempa-jabe, line 892: Missing tab

gempa-jabe:
I am fine with adding tooltips to each of the columns. What is your actual intention for that?

luca-s:
In my use case the column (an origin ID and a long text in another case) is too long to be visualised without horizontally scrolling the events table. This is because the table is quite packed with important information already and I do not want to hide any of them. With the tooltip I do not need to do so, I can simply leave the mouse at the beginning of the custom column, which is only partially visible and see the content in the tooltip.

gempa-jabe:
Thank you for clarification. To accept the PR you need to sign the CLA. If you agree just contact me otherwise close the PR and I will apply the changes myself according to your description.

luca-s:
@gempa-jabe sure, no problem for signing the CLA, I'll contact you via email.

luca-s:
@gempa-jabe

> otherwise close the PR and I will apply the changes myself according to your description.

Just for this one, do you mind  applying the patch yourself? thank you very much, I am closing the PR  now

PR #13, opened 2021-01-18
Fix gap interpolation

damb:
This PR fixes two minor issues regarding gap interpolation:

- Correctly return whether gap was handled
- Compare `_gapThreshold` with sampling interval; the minimum allowed gap threshold is twice the sampling interval

### DIFF
--- a/libs/seiscomp/processing/waveformprocessor.cpp
+++ b/libs/seiscomp/processing/waveformprocessor.cpp
@@ -312,6 +312,16 @@ bool WaveformProcessor::store(const Record *record) {
 	// NOTE: Do not use else here, because lastRecord can be set nullptr
 	//       when calling reset() in handleGap(...)
 	if ( !_stream.lastRecord ) {
+
+		const auto minGapThres = 2 * 1.0 / record->samplingFrequency();
+		if ( minGapThres > _gapThreshold ) {
+			SEISCOMP_WARNING("Gap threshold smaller than twice the sampling interval: "
+			                 "%fs < %fs. Resetting gap threshold.",
+			                 static_cast<double>(_gapThreshold), minGapThres);
+
+			_gapThreshold = minGapThres;
+		}
+
 		try {
 			initFilter(record->samplingFrequency());
 		}
--- a/libs/seiscomp/processing/waveformprocessor.cpp
+++ b/libs/seiscomp/processing/waveformprocessor.cpp
@@ -438,9 +438,8 @@ bool WaveformProcessor::handleGap(Filter *filter, const Core::TimeSpan& span,
 				double value = lastSample + di*delta;
 				fill(1, &value);
 			}
+			return true;
 		}
-
-		return true;
 	}
 
 	return false;


damb:
Apparently, the CLA here is not the same as in [SeisComP/main](https://github.com/SeisComP/main). @gempa-jabe, could you add me, please? Thanks.

gempa-jabe:
The CLA here is between gempa Gmbh and you while the CLA in main is between GFZ and you. This is due to split copyrights. You need to sign https://github.com/gempa/cla/blob/master/icla.md as well. Sorry for confusion. You don't need to sign for any other CLAs again within this repository collection.

gempa-jabe:
Thank you for signing the CLA - @cla-bot check

gempa-jabe:
I agree with the return value. With respect to the gap minimum threshold, why did you chose 2 * dt? We have decided to use 0.5 * dt as default threshold for accepting time gaps, see RecordSequence or RecordPolyline. Actually all code places in which traces are being formed from record lists use that value. Would it break something at your end to also apply the value here?

gempa-jabe:
Sorry, my bad. A value < 2 does not make sense for interpolation.

libs/seiscomp/processing/waveformprocessor.cpp
```diff
@@ -284,6 +284,15 @@ bool WaveformProcessor::store(const Record *record) {
 	if ( _stream.lastRecord ) {
 		if ( record == _stream.lastRecord ) return false;
 
+		const auto minGapThres = 2 * 1.0 / _stream.fsamp;
```
gempa-jabe, line 287: As this check only depends on _stream.fsamp it would be more efficient to only check for it when _stream.fsamp is being set. That is done in `initFilter`. Would you agree?

libs/seiscomp/processing/waveformprocessor.cpp
```diff
@@ -284,6 +284,15 @@ bool WaveformProcessor::store(const Record *record) {
...
+		if (minGapThres > _gapThreshold) {
```
gempa-jabe, line 288: Please use spaces between parenthesis, e.g. `if ( minGapThres > _gapThreshold )`.

libs/seiscomp/processing/waveformprocessor.cpp
damb: That's true. Though, then the naming of the `initFilter` method is kind of misleading. So I would do it before this block here https://github.com/SeisComP/common/blob/0932f584e44c600d84cfded3ecf4f0f8affbf7b9/libs/seiscomp/processing/waveformprocessor.cpp#L315-L322

damb: On the other hand, what happens if the sampling frequency changes?:
* filters should be reinitialized
* the gap threshold should be checked once again

gempa-jabe: Yes, `initFilter` should actually be called `initStream` ;) You can also do it before this block. Changing sampling rates are not handled currently. Once this is implemented, a new function `setupStream` should be called.

gempa-jabe:
Thanks a lot. You also got the spacings right ;)

damb:
Do you want me to rebase?

damb:
> Thanks a lot. You also got the spacings right ;)

Do you have something like a `.clang_format`? I mean, for future PRs.

gempa-jabe:
> Do you want me to rebase?

No, Github can do that for us.

gempa-jabe:
> Do you have something like a .clang_format? I mean, for future PRs.

Unfortunately not. We just have a basic document for coding conventions.
https://github.com/SeisComP/seiscomp/blob/master/doc/base/coding-conventions.rst

damb:
> Unfortunately not. We just have a basic document for coding conventions.
> https://github.com/SeisComP/seiscomp/blob/master/doc/base/coding-conventions.rst

Ok. I've seen this already. Thanks.

gempa-jabe:
Can you please explain why you came up with this sampling rate fix? Did you run into issues?

I see a potential problem when the sampling rate changes from e.g. 1 sps to 20 sps. Then the minimum gap threshold will be set to 2s and not set back to 0.1s when the sampling rate changes and when the initial gap threshold was set to a value lower than 2s.

damb:
> Can you please explain why you came up with this sampling rate fix? Did you run into issues?

I stumpled over this issue, while creating an object of `Seiscomp::Processing::WaveformProcessor` with an `gapThreshold` equal to zero. Then, all gaps, i.e. also sub-sampling interval gaps were handled. In fact, I had a case with records arriving with tiny gaps of a few microseconds.

> I see a potential problem when the sampling rate changes from e.g. 1 sps to 20 sps. Then the minimum gap threshold will be set to 2s and not set back to 0.1s when the sampling rate changes and when the initial gap threshold was set to a value lower than 2s.

That's right. When implementing the feature *handle changing sampling rates* this must be taken into account.

PR #15, opened 2021-03-12
Save last sample before filling values

damb:
If I'm not completely mistaken, there is an issue when saving the last sample after filling (filtering) the data. Instead,
the raw sample before filling needs to be saved. This is required since gaps are handled before data is filled.


gempa-jabe:
Thanks a lot, well spotted.

PR #16, opened 2021-03-31
Fix typo

damb:
**Changes**:
- Fix typo (see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Lanczos_resampling))


gempa-jabe:
Thanks a lot.

PR #21, opened 2021-05-26
Support SQLite in-memory DB

damb:
**Features and Changes**:
- Support [SQLite in-memory DB](https://sqlite.org/inmemorydb.html) (using the DB URL `sqlite3://:memory:`)

### DIFF
--- a/plugins/database/sqlite/sqlitedatabaseinterface.cpp
+++ b/plugins/database/sqlite/sqlitedatabaseinterface.cpp
@@ -59,17 +59,19 @@ SQLiteDatabase::~SQLiteDatabase() {
 
 // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 bool SQLiteDatabase::open() {
-	std::string filename = Environment::Instance()->absolutePath(_host);
+	std::string uri(_host);
+	if ( uri != ":memory:" ) {
+		uri = Environment::Instance()->absolutePath(_host);
 
-	FILE *fp = fopen(filename.c_str(), "rb");
-	if ( fp == NULL ) {
-		SEISCOMP_ERROR("databasefile '%s' not found", filename.c_str());
-		return false;
+		FILE *fp = fopen(uri.c_str(), "rb");
+		if ( fp == NULL ) {
+			SEISCOMP_ERROR("databasefile '%s' not found", uri.c_str());
+			return false;
+		}
+		fclose(fp);
 	}
 
-	fclose(fp);
-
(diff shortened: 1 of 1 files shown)


gempa-jabe:
That additions makes sense, thank you. Although I cannot think of a use-case. What do you need it for?

jsaul:
For instance I use sqlite throw-away databases to process XML dumps if direct XML processing is not an option or not convenient. To populate the database I use scdb but this is crazy slow if the database is on a harddisk. A ramdisk is orders of magnitudes faster and I presume that sqlite3://:memory will be even faster.

Bottom line is yes, there are use cases and the contribution is highly appreciated. Thanks a lot!

damb:
An exemplary use case: A module which requires `EventParameters` as input parameters (usually provided by `--ep`) could import the `*.scml` into an in-memory SQLite DB (once, the DB is initialized with the SeisComP DB schema). The module then is able to make use of `SeisComP::DataModel` facilities such as e.g. `Seiscomp::DataModel::DatabaseQuery` which simplifies the `EventParameter` handling tremendously (e.g. retrieve the `DataModel::Event` to a corresponding `DataModel::Origin` `publicID`).

Though, at the moment bringing the SeisComP DB schema to the DB is still quite *hacky* (e.g. by means of parsing the SQL DDL from `$SEISCOMP_ROOT/share/db/*.sql`). It would be nice if there would be some kind of cleaner solution.

gempa-jabe:
> A ramdisk is orders of magnitudes faster and I presume that sqlite3://:memory will be even faster.

scdb won't benefit from this change as the database is gone afterwards. Creating and using the database temporarily during runtime makes sense although I prefer to traverse the EventParameters and work with the objects in memory. I presume that a few helper functions would do the job as well and the input XML is not huge. Adding a hard dependency of SQLite to your module is the consequence then. Anyway, great addition.

jsaul:
OIC. It only exists during runtime of one program? That limits the number of use cases, although a Python program can be very long. One would then need the scdb functionality in Python.

What I sometimes do in a shell script:

```
$ template=/dev/shm/seiscomp.db-template
$ sqlite3 $template ".read $HOME/seiscomp/share/db/sqlite3.sql"
$ scdb -i journal.xml -d sqlite3://$template
```

Here I pre-populated the database using a SC journal XML dump.

And then I keep this template database and just copy it whenever I need a clean database with journal. And/or inventory, config etc.

But perhaps this use case is not as straightforward to implement using the in-memory DB than I initially hoped.

jsaul:
We need more SeisComP/Python hackers!

damb:
> But perhaps this use case is not as straightforward to implement using the in-memory DB than I initially hoped.

This is not possible. See also: https://stackoverflow.com/a/15720872/12083893.

PR #22, opened 2021-05-27
Allow TimeWindowBuffer to return its TimeWindow

luca-s:
This PR adds a convenience method that returns the TimeWindow the TimeWindowBuffer is constructed with. Currently there is no way to know that TimeWindow. In my code I have to keep a separate data structure just to access that information. Thanks for your review.

### DIFF
--- a/libs/seiscomp/core/recordsequence.h
+++ b/libs/seiscomp/core/recordsequence.h
@@ -175,6 +175,8 @@ class SC_SYSTEM_CORE_API TimeWindowBuffer : public RecordSequence {
 		//! Return percentage of available data within TimeWindowBuffer's TimeWindow
 		double availability() const;
 
+		//! Return the TimeWindow used in constructing the TimeWindowBuffer
+		const TimeWindow& timeWindow() const;
 
 	// ----------------------------------------------------------------------
 	//  Members
@@ -252,6 +254,9 @@ inline size_t RecordSequence::recordCount() const {
 	return size();
 }
 
+inline const TimeWindow& TimeWindowBuffer::timeWindow() const {
+	return _timeWindow;
+}
 
 }
 
--- a/libs/seiscomp/core/recordsequence.h
+++ b/libs/seiscomp/core/recordsequence.h
@@ -175,8 +175,8 @@ class SC_SYSTEM_CORE_API TimeWindowBuffer : public RecordSequence {
 		//! Return percentage of available data within TimeWindowBuffer's TimeWindow
 		double availability() const;
 
-		//! Return the TimeWindow used in constructing the TimeWindowBuffer
-		const TimeWindow& timeWindow() const;
+		//! Return the buffered TimeWindow
+		const Core::TimeWindow &bufferedTimeWindow() const;
 
 	// ----------------------------------------------------------------------
 	//  Members
@@ -227,6 +227,11 @@ class SC_SYSTEM_CORE_API RingBuffer : public RecordSequence {
 		//! clear the buffer
 		void reset() { clear(); }
 
+		//! Return the maximum number of records the RingBuffer stores
+		unsigned int maxNumRecords() const;
+	
+		//! Return the TimeSpan the RingBuffer stores
+		const Core::TimeSpan &timeSpan() const;
 
 	// ----------------------------------------------------------------------
 	//  Members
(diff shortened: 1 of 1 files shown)


damb:
@luca-s, have you seen this?: https://github.com/SeisComP/common/blob/02c0e2854199ac64bdd5ed862306be4d9a637bfa/libs/seiscomp/core/recordsequence.h#L99-L100

Doesn't this result in a conflict?

libs/seiscomp/core/recordsequence.h
```diff
@@ -175,6 +175,8 @@ class SC_SYSTEM_CORE_API TimeWindowBuffer : public RecordSequence {
 		//! Return percentage of available data within TimeWindowBuffer's TimeWindow
 		double availability() const;
 
+		//! Return the TimeWindow used in constructing the TimeWindowBuffer
+		const TimeWindow& timeWindow() const;
```
gempa-jabe, line 179: Please replace `const TimeWindow& timeWindow() const` with `const TimeWindow &timeWindow() const` for consistency.

libs/seiscomp/core/recordsequence.h
```diff
@@ -252,6 +254,9 @@ inline size_t RecordSequence::recordCount() const {
 	return size();
 }
 
+inline const TimeWindow& TimeWindowBuffer::timeWindow() const {
```
gempa-jabe, line 257: See my other request.

gempa-jabe (changes requested):

gempa-jabe:
For consistency also method to retrieve the TimeSpan and the maximum number of records should be added to RingBuffer. And note that `TimeWindowBuffer::timeWindow` will overload `RecordSequence::timeWindow`. Another name should be found for that.

luca-s:
> @luca-s, have you seen this?:
> 
> https://github.com/SeisComP/common/blob/02c0e2854199ac64bdd5ed862306be4d9a637bfa/libs/seiscomp/core/recordsequence.h#L99-L100
> 
> Doesn't this result in a conflict?

Ops, my bad. Yes, I need to rename the method

luca-s:
> For consistency also method to retrieve the TimeSpan and the maximum number of records should be added to RingBuffer. And note that `TimeWindowBuffer::timeWindow` will overload `RecordSequence::timeWindow`. Another name should be found for that.

Sounds good, I will do that

libs/seiscomp/core/recordsequence.h
```diff
@@ -175,6 +175,8 @@ class SC_SYSTEM_CORE_API TimeWindowBuffer : public RecordSequence {
 		//! Return percentage of available data within TimeWindowBuffer's TimeWindow
 		double availability() const;
 
+		//! Return the buffered TimeWindow
```
gempa-jabe, line 178: To be pedantic, this is actually not the buffered time window. The buffered time window is returned with `RecordSequence::timeWindow`. I am also not sure about a better name, maybe `capacity`. But note that the actual buffered time window can be larger than the requested time window because records may overlap the start and/or end time.

libs/seiscomp/core/recordsequence.h
```diff
@@ -225,6 +227,11 @@ class SC_SYSTEM_CORE_API RingBuffer : public RecordSequence {
 		//! clear the buffer
 		void reset() { clear(); }
 
+		//! Return the maximum number of records the RingBuffer stores
+		unsigned int maxNumRecords() const;
```
gempa-jabe, line 231: Please rename to `numberOfRecords` to be consistent with other interfaces.

libs/seiscomp/core/recordsequence.h
luca-s: "To be pedantic", you have to be, the API needs to be clear, I understand that. Maybe it was the naming issue that prevented the implementation from providing a method for that ;) What about `sequenceTimeWindow`?

gempa-jabe: It is really difficult. Actually the RingBuffer suffers as well from correct naming. When I think about `numberOfRecords` one might expect the number of records currently held by the sequence. What is required is a naming prefix that gives access to the configuration and is consistent with both sequence implementations. Just a proposal:

```
configTimeWindow()
configNumberOfRecords()
configTimeSpan()
```

I am not sure if I like that ;) For the number of records `maxRecords` is probably the best but that does not work for the other attributes because of the overlap. Otherwise someone could complain that `maxTimeWindow() < timeWindow()`. That does not follow your expectation.

luca-s: I agree. Let's think about the naming  and then I will be back with a proposal

luca-s: And here is the proposal. What about: 
```
timeWindowToStore()
numberOfRecordsToStore()
timeSpanToStore()

```

gempa-jabe: I can't come up with a better suggestion so please go ahead.

luca-s: Thanks, I updated the PR

gempa-jabe:
I completely missed the erroneous member implementation of `inline const Core::TimeWindow &timeWindowToStore() const`. I haven't compiled your branch before merging and neither have you ;) I just fixed it.

PR #26, opened 2021-06-01
Consistently reset filters

damb:
**Features and Changes**:
- Consistently reset the filter state when (re)configuring the sampling frequency. (Otherwise, calling `Seiscomp::Processing::WaveformProcessor::initFilter()` in `Seiscomp::Processing::WaveformProcessor::setupStream()` may leave filters with state in case of changing sampling frequencies.)
- Consistently apply SeisComP coding conventions with regards to virtual member functions and overrides.

Please review carefully.

### DIFF
--- a/libs/seiscomp/math/filter/iirdifferentiate.cpp
+++ b/libs/seiscomp/math/filter/iirdifferentiate.cpp
@@ -30,15 +30,14 @@ namespace Filtering
 
 template <typename T>
 IIRDifferentiate<T>::IIRDifferentiate(double fsamp) : _fsamp(fsamp) {
-	reset();
 	setSamplingFrequency(fsamp);
 }
 
 
 template <typename T>
 IIRDifferentiate<T>::IIRDifferentiate(const IIRDifferentiate<T> &other) {
-	_fsamp = 0.0;
-	reset();
+  _fsamp = 0.0;
+  reset();
 }
 
 
@@ -51,7 +50,10 @@ void IIRDifferentiate<T>::reset() {
 
 template <typename T>
 void IIRDifferentiate<T>::setSamplingFrequency(double fsamp) {
+	if ( fsamp == _fsamp) return;
+
 	_fsamp = fsamp;
+  reset();
 }
 
 
--- a/libs/seiscomp/math/filter/iirintegrate.cpp
+++ b/libs/seiscomp/math/filter/iirintegrate.cpp
@@ -29,7 +29,8 @@ namespace Filtering
 
 
 template <typename T>
-IIRIntegrate<T>::IIRIntegrate(double a, double fsamp) {
+IIRIntegrate<T>::IIRIntegrate(double a, double fsamp)
+{
 	init(a);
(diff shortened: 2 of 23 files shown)


libs/seiscomp/math/filter/iirdifferentiate.cpp
```diff
@@ -30,15 +30,13 @@ namespace Filtering
...
-	reset();
 	setSamplingFrequency(fsamp);
 }
 
 
 template <typename T>
 IIRDifferentiate<T>::IIRDifferentiate(const IIRDifferentiate<T> &other) {
-	_fsamp = 0.0;
-	reset();
+  setSamplingFrequency(other._fsamp);
```
gempa-jabe, line 39: If the compiler initializes _fsamp with 0.0 then this function does nothing and the other reset states are still unconfigured. The original two lines stable in that regard.

libs/seiscomp/math/filter/iirintegrate.cpp
```diff
@@ -71,9 +71,13 @@ void IIRIntegrate<T>::setSamplingFrequency(double fsamp) {
 
 	double dt = 1./fsamp;
 
-	_a0 = _ia0 * dt;
+  double a0 = _ia0 * dt;
```
gempa-jabe, line 74: Does this block intent to check if the sampling frequency has changed? Why not just checking it at the top of the function as with all other filters?

libs/seiscomp/math/filter/taper.cpp
```diff
@@ -34,6 +34,17 @@ InitialTaper<TYPE>::InitialTaper(double taperLength, TYPE offset, double fsamp)
 		setSamplingFrequency(fsamp);
 }
 
+template<typename TYPE>
+void InitialTaper<TYPE>::setSamplingFrequency(double fsamp) {
+	if (_samplingFrequency == fsamp) return;
```
gempa-jabe, line 39: Please add spaces between parenthesis: `if ( a )`

gempa-jabe (changes requested):

gempa-jabe:
Thanks a lot for the work on this. Cleaning up the filter code has been on the list for some time.

libs/seiscomp/math/filter/iirintegrate.cpp
damb: Yes. I didn't like it, neither. Instead, it is much cleaner to store the sampling frequency as a member variable and do the check as with all other filters.

gempa-jabe: OK, I see. Adding _fsamp will break binary compability. So this might be the best approach.

damb: I went for the original attempt with a minor change.

>  Adding _fsamp will break binary compability.

Again, I learned something :) ...

gempa-jabe:
I have checked the code now multiple times and it should not break anything. I am going to merge it. Thanks again.

PR #27, opened 2021-06-02
scolv/ focal sphere/ quadrants colours : darker?

FMassin:
As discussed in https://forum.seiscomp.de/t/scv4-scolv-focal-sphere-quadrants-colours , the colour of the unconfirmed focal sphere compression quadrants is a bit confusing. Keeping the same kind of colour than original, there are 3 other possibilities:  « Dark », « Mid » and « Midlight » (see attached). I propose to use « Mid ». 

I understand the colour for the quadrants after being confirmed is « Highlight » with a 25% blend with background color, which is good, and I don't intend to propose any change on that.

Capture d’écran 2021-06-02 à 15 54 53

### DIFF
--- a/libs/seiscomp/gui/datamodel/originlocatorview.cpp
+++ b/libs/seiscomp/gui/datamodel/originlocatorview.cpp
@@ -1102,7 +1102,7 @@ class PlotWidget : public OriginLocatorPlot {
 			Math::tensor2matrix(_tensor, tmp);
 
 			_renderer.setPColor(palette().color(QPalette::Base));
-			_renderer.setTColor(palette().color(QPalette::AlternateBase));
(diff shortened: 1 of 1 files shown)


gempa-jabe:
@cla-bot check

gempa-jabe:
"Mid" is actually looking nice as well. I am fine with that change.

FMassin:
I think so too :) Here's more screenshots...

Capture d’écran 2021-06-02 à 18 25 43
Capture d’écran 2021-06-02 à 18 24 41
Capture d’écran 2021-06-02 à 18 24 52

gempa-jabe:
Then let's merge it. Thanks.

FMassin:
Great! Thanks.

PR #30, opened 2021-06-07
Correctly parse `resample` RecordStream URL

damb:
**Changes**:
- Correctly parse the `resample` RecordStream URL (in particular, do not discard optional parameters)

I hope I got the [documentation](https://www.seiscomp.de/doc/apps/global_recordstream.html#rs-resample) right. Note that the `dec` RecordStream seems to have similar issues. If required, I'm going to fix that, too.

### DIFF
--- a/libs/seiscomp/io/recordstream/resample.cpp
+++ b/libs/seiscomp/io/recordstream/resample.cpp
@@ -80,17 +80,13 @@ bool Resample::setSource(const string &source) {
 	}
 
 	string name = source;
+	string service = name.substr(0, pos);
 	string addr = name.substr(pos+1);
-	name.erase(pos);
 
-	string service;
-
-	pos = name.find('?');
-	if ( pos == string::npos )
-		service = name;
-	else {
-		service = name.substr(0, pos);
-		name.erase(0, pos+1);
+	pos = addr.find('?');
+	if ( pos != string::npos ) {
+		name = addr.substr(pos+1);
+		addr.erase(pos);
 
 		vector<string> toks;
 		Core::split(toks, name.c_str(), "&");


gempa-jabe:
Can you please give an example of an URL that did not work and works now?

damb:
An URL such as e.g. `resample://fdsnws/eida-federator.ethz.ch/fdsnws/dataselect/1/query?rate=20` did not configure the correct `_targetRate`. Instead, the default target rate was used. (AFAIU, basically all optional parameters related to the `resample` RecordStream were omitted and passed to the *proxy record stream*, instead.) With this PR they should be configured correctly.

gempa-jabe:
Ah, I see. But that is not how the URL should work for this particular service. It is intended to look like this:

```
resample://fdsnws?rate=20/eida-federator.ethz.ch/fdsnws/dataselect/1/query
```

Otherwise you are not able to append parameters to the proxy stream. I know this seems a bit ugly but there is no other option currently.

damb:
Ok, thanks for the clarification, @gempa-jabe. Now, the implementation becomes clear. Also, with your comment this PR is obsolete.

Anyway, I think the [documentation](https://www.seiscomp.de/doc/apps/global_recordstream.html#rs-resample) could describe the usage and configuration more in detail. E.g. something like

```
resample://proxy-service[?resample-parameters]/proxy-address[?proxy-parameters]
```

where
  - `proxy-service` is the proxy *service identifier prefix* of the *proxy stream* (e.g. `slink`)
  - `resample-parameters` are configuration parameters which specify the configuration of the `resample` RecordStream (e.g. `rate=20`) (optional)
  - `proxy-address` refers to the address of the *proxy stream* (e.g. `localhost:18000`)
  - `proxy-parameters` are configuration parameters which specify the configuration of the *proxy stream* (e.g. `timeout=60&retries=5`) (optional).

Particularly, the URI in the *Examples* section is confusing (i.e. `service://source?options`).

gempa-jabe:
We will address this.

damb:
What a coincidence :+1:: 914d6cd88a3134840a180b07f4ba9e49471050bc

PR #32, opened 2021-07-15
Update mysql.sql

filefolder:
change normalizationFactor to just DOUBLE as alluded to here: https://github.com/SeisComP/main/issues/27#issuecomment-876337367

(older sensors can typically have negative A0 factors)


gempa-jabe:
I have pushed a commit that removes signedness of the two attributes `normalizationFactor` and `normalizationFrequency`. Both MYSQL and SQLite scripts have changed.

PR #34, opened 2021-08-11
Fix source code documentation typos

damb:
- Fixes some typos in the source code documentation


gempa-jabe:
Thanks for changing all this. Please note that the code and documentation is generated. So I have to regenerate the code and fix the typo in the input. That is actually true for all data model classes.

damb:
Ok, thanks for letting me know. Should I adopt this PR for fixing just the typos which do not relate to `DataModel` stuff? If I understand you correctly, you're going the `DataModel` related typos by yourself.

gempa-jabe:
Yes, I am going to fix the typos. They also have effect on the API HTML documentation which is also generated. When I have pushed the changed then you can just rebase your branch and I will pull all the other fixes from you. That is probably the easiest way.

gempa-jabe:
Changes have been pushed ...

gempa-jabe:
Perfect, thanks.

PR #36, opened 2021-08-13
Sync safety time window

damb:
**Changes**:
- Only accept margins >= 0
- Also sync the `_safetyTimeWindow` member

### DIFF
--- a/libs/seiscomp/processing/timewindowprocessor.cpp
+++ b/libs/seiscomp/processing/timewindowprocessor.cpp
@@ -18,6 +18,7 @@
  ***************************************************************************/
 
 
+#include <seiscomp/core/exceptions.h>
 #include <seiscomp/processing/timewindowprocessor.h>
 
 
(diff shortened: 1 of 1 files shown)


gempa-jabe:
Not sure what should happen in the other case when safetyMargin is less than 0. I think throwing an exception (as there is no return type) is appropriate.

damb:
> Not sure what should happen in the other case when safetyMargin is less than 0. I think throwing an exception (as there is no return type) is appropriate.

You mean e.g. `Core::ValueException`?

gempa-jabe:
I would prefer either GeneralException or even UnderflowException. ValueException is used if a value is not set but should be read, e.g. origin.creationInfo().

damb:
One additional remark which is somehow OT. Shouldn't a `Processing::TimeWindowProcessor` also take the [_initTime](https://github.com/SeisComP/common/blob/dd4ebc710f8bfd9287c0064c7a4e492d452c74d0/libs/seiscomp/processing/waveformprocessor.h#L389) into account?

gempa-jabe:
`_initTime` is used for continuous processing to initialize filters and stuff. In `TimeWindowProcessor` this is done via the safetyMargin. I see your point. So it should (if ever set) also go into the time window and might be redundant to `safetyMargin`.

damb:
Well, AFAIU `_initTime` and `_safetyMargin` serve different purposes: as already mentioned `_initTime` is mainly used for filter initialization, while `_safetyMargin` is used for e.g. SNR analysis. I don't think they are redundant. Therefore, I propose to actually add the `_initTime` to the `_safetyMargin`.

gempa-jabe:
Add it to `_safetyMargin` or better include it into the `safetyTimeWindow` computation? I think the latter is more appropriate.

Actually the `safetyMargin` was introduced for filter initialization and to have some more data available than required. But this data is optional. If you need additional data for SNR analysis then adding them to the desired time window (which is mandatory) would be correct, wouldn't it? Otherwise you end up sometimes with 10 seconds of additional data, sometime 30 seconds depending on data availability. But maybe this is OK for your processing.

damb:
> Add it to `_safetyMargin` or better include it into the `safetyTimeWindow` computation? I think the latter is more appropriate.
> 
> Actually the `safetyMargin` was introduced for filter initialization and to have some more data available than required. But this data is optional. If you need additional data for SNR analysis then adding them to the desired time window (which is mandatory) would be correct, wouldn't it? Otherwise you end up sometimes with 10 seconds of additional data, sometime 30 seconds depending on data availability. But maybe this is OK for your processing.

In this case including is more appropriate. I agree. But then, in future `safetyTimeWindow` also could (and should) be removed from the API, shouldn't it?

gempa-jabe:
`_initTime` has been included ...

damb:
Perfect :+1:. Thanks.

PR #43, opened 2021-10-27
New Split RecordStream

luca-s:
Dear developers,

this is a proposal for the inclusion of a new "Split" RecordStream into SeisComP. The purpose of this RecordStream is to split the requests across several services accordingly to user selected rules.

```
The URL format is the following:
 split://type1/source1??match=pattern;type2/source2??match=pattern;...;typeN/sourceN??match=pattern
where:
  sourceN is either source or (source)
  pattern is NET.STA.LOC.CHA and the special charactes ? * | ( ) are allowed
```

E.g.
```
split://slink/host1??8D.*.*.*;slink/host2??2C.*.*.*;slink/host3??*.*.*.*
split://slink/host1??(8D|4D).*.*.*;slink/host2??*.*.*.*
split://slink/host1??*.*.*.(HH|EH)?;slink/host2??*.*.*.*
split://combined/(server1:18000;server1:18001)??CH.*.*.*;combined/(server2:18000;server2:18001)??2C.*.*.*
```

If you are interested in the integration of this feature into SeisComP I would like to ask you to review the current URL format and the RecordStream name. 

Also I wonder if it would be better to achieve the same goal by adding a special option to `balanced` RecordStream, since the code of `balanced` and `split` is 90% the same.

E.g.
```
balanced://slink/host1??splitRule=(8D|4D).*.*.*;slink/host2??splitRule=*.*.*.*
# a missing splitRule defaults to *.*.*.*
balanced://slink/host1??splitRule=(8D|4D).*.*.*;slink/host2
```


luca-s:
Here are some improvements:

* Fixed concurrency bugs in `BalancedConnection` + minor issues
* Renamed Split to Routing RecordStream
* `RoutingConnection` class now inherits from `BalancedConnection`  so that there is no concurrent code duplication

I would still like to have a common base class between `RoutingConnection` and `BalancedConnection`  instead of inherit from `BalancedConnection` . Even better would be to add the option `??splitRule=net.sta.loc.cha` to the `balanced` recordstream and avoid a new class entirely.

luca-s:
@gempa-jabe how should we proceed with this PR?

gempa-jabe:
The code you have reworked looks pretty good. The implementation of the routing recordstream is simple and easy to understand. I agree that it is desirable to have a base class for both balanced and routing to do the actually threading and synchronization. It would be easier to create new implementations based on that approach. I don't like the idea of a complex implementation that supports everything. Having well separated code units is much more appealing to me. ;)

luca-s:
I pushed another commit where I created a common base class for both "balanced" and "routing" RecordStream.

gempa-jabe:
Yes, that is looking good. The concrete implementations are pretty compact. Some details:

* Concurrent should be in the public namespace and not private as it should serve as a base class for future implementations. I would prefer it under the namespace `Seiscomp::RecordStream`.
* Classes in a "private" namespace should not expose its interface in a public header. Otherwise users might want to instantiate them explicitly in their code but obviously that is not intended by the developer. I would include the header into the implementation and implement everything in a real private namespace `namespace {`. That way the implementation is hidden, does not pollute the namespace and changes to the class memory layout will not break the ABI.

I don't expect you to do those changes, I can do it. I would like to confirm with @andres-h if the current implementation is OK with him as he is the author of the balanced recordstream.

luca-s:
I don't want to put any pressure, just double checking with you that you are not waiting anything from my side. The ball is now on your court right?

gempa-jabe:
You are right. I would like @andres-h to check the code as it would possibly break the balanced record stream. @andres-h, would you do that?

andres-h:
Yes, sorry I overlooked this. I will check tomorrow.

andres-h:
I tested the new balanced recordstream and it works as expected.

(Looks like "delete thread" is still missing.)

luca-s:
> (Looks like "delete thread" is still missing.)

ah, your are right. I'll push a fix.

luca-s:
I switched from list of thread pointers to list of threads

gempa-jabe:
I am going to pull all your changes into a local branch, rebase it, do the modifications I was talking about and test it.

gempa-jabe:
I will close it as it is being replaced with #45.

PR #45, opened 2021-12-16
Add routing recordstream

gempa-jabe:
@luca-s, @andres-h this is my rework of your proposed implementation. Please take a look. What I did was:

* Replace std::regex with boost:regex to support gcc 4.5
* Expose RcordStream::Concurrent as abstract public class to be used by other implementations later
* Implement BalancedConnection and RoutingConnection in private namespaces to not pollute the public namespace
* Remove the headers from being installed as they are private implementations
* Clean up headers and types

Both `balanced://` and `routing://` should still work as expected. If you could please give it a try at your end.


gempa-jabe:
This replaces #43.

luca-s:
Thanks I will have a look and test it

luca-s:
@gempa-jabe did you drop on purpose the constructors of `RoutingConnection` and `BalancedConnection` that accept a URL parameter?

gempa-jabe:
Yes, the implementation is private an cannot be instantiated from outside. So the constructor cannot be called. Only the factory will ever call the default constructor.

andres-h:
Looks OK to me and balanced works as expected.

luca-s:
I found an issue on the routing, it's an implementation difference between `std` and `boost` `regex_replace`.  The boost version needs an addition flag `match_flag_type::format_literal`

image

gempa-jabe:
I have forced pushed the code and included the additional flag.

luca-s:
LGTM, I didn't find other issues.Thank you @gempa-jabe

gempa-jabe:
Great, then I merge it. Thanks to both of you.

luca-s:
Sure, I had already thought about it and then I forgot

PR #47, opened 2022-01-10
Mark overridden virtual methods with `override`

damb:


gempa-jabe:
Thanks for adding those specifiers. Actually there are still some more places which require this modification. We internally agreed to do the transition to C++11 step by step either when the compiler starts to complain or when we touch the file anyway.

Would you mind to prefix the commit message with the "[core]"?

Issue #11, opened 2021-01-06
Application cleanup for inventory singleton


damb:
Hi devs,

while implementing unittests for a concrete implementation of `Seiscomp::Client::StreamApplication` I run into the issue that the `inventory_` member of the `Seiscomp::Client::Inventory` singleton instance i.e. https://github.com/SeisComP/common/blob/ecfeba10d3bf961d8ac426aeedf1a8dc07e453fd/libs/seiscomp/client/inventory.h#L146 is not properly reset during application cleanup. Of course, I could add something like
```C++
void MyApplication::done() {
   ...
   Seiscomp::Client::Inventory::Instance()->setInventory(nullptr);
}
```
which would do the job. However, IMO it would be cleaner to add this line to https://github.com/SeisComP/common/blob/ecfeba10d3bf961d8ac426aeedf1a8dc07e453fd/libs/seiscomp/client/application.h#L465 since `Seiscomp::Client::Application` handles the `Seiscomp::Client::Inventory`, anyway.

What do you think?

EDIT: Correct namespace

gempa-jabe:
That sounds reasonable. Actually since Inventory is a global instance (outside of Application scope) it is also arguable if the application should clean up the state of this global instance. The application loads the global inventory based on some rules but that's it. I don't know of any implementation which uses the inventory after the application has been destroyed so adding that cleanup is OK.

gempa-jabe:
See 04050c85931b2c6c4b798f05c60f21051c226954.

gempa-jabe:
Please let me know if that fixes your tests.

damb:
Thanks, @gempa-jabe. The fix does its job and works correctly.

PR #49, opened 2022-01-11
[client] Reset config at app shutdown

damb:
In accordance to #11.

### DIFF
--- a/libs/seiscomp/client/application.cpp
+++ b/libs/seiscomp/client/application.cpp
@@ -1047,8 +1047,16 @@ bool Application::reloadBindings() {
 				return false;
 			}
 		}
+		else {
+			SEISCOMP_ERROR("Failed to load configuration module");
+			return false;
+		}
 
 		DataModel::Config* config = ConfigDB::Instance()->config();
+		if ( !config ) {
+			return false;
+		}
+
 		for ( size_t i = 0; i < config->configModuleCount(); ++i ) {
 			if ( config->configModule(i)->name() == _configModuleName ) {
 				_configModule = config->configModule(i);
@@ -1242,6 +1250,7 @@ void Application::done() {
 	_database = nullptr;
 
 	Inventory::Instance()->setInventory(nullptr);
+	ConfigDB::Instance()->setConfig(nullptr);
 
 	SEISCOMP_DEBUG("Leaving ::done");
 }
--- a/libs/seiscomp/client/configdb.cpp
+++ b/libs/seiscomp/client/configdb.cpp
@@ -262,6 +262,9 @@ void ConfigDB::load(const char *xml) {
 	ar.close();
 }
 
+void ConfigDB::setConfig(DataModel::Config *config) {
+	_config = config;
+}
 
 Seiscomp::DataModel::Config* ConfigDB::config() {
 	return _config.get();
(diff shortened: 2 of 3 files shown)


libs/seiscomp/client/application.cpp
```diff
@@ -1046,9 +1046,16 @@ bool Application::reloadBindings() {
 				SEISCOMP_ERROR("No database query object");
 				return false;
 			}
+		} else {
```
gempa-jabe, line 1049: Please move `else` to a new line, see above for an example.

libs/seiscomp/client/configdb.cpp
```diff
@@ -262,6 +262,9 @@ void ConfigDB::load(const char *xml) {
 	ar.close();
 }
 
+void ConfigDB::setConfig(DataModel::Config* config) {
```
gempa-jabe, line 265: Please attach the asterisk to `config`,  `DataModel::Config *config`.

libs/seiscomp/client/configdb.h
```diff
@@ -52,7 +52,8 @@ class SC_SYSTEM_CLIENT_API ConfigDB {
 			const std::set<std::string>& parameterNames = std::set<std::string>());
 
 		void load(const char *xml);
-
+		
+		void setConfig(DataModel::Config* config);
```
gempa-jabe, line 56: Please attach the asterisk to config, see other comment.

gempa-jabe (changes requested):

gempa-jabe:
Great, thanks.

damb:
Note that with this PR, modules not explicitly requiring a loading the config module must call `setLoadConfigModuleEnabled(false)`. This is the default, anyway https://github.com/SeisComP/common/blob/df83362fbb6ea29eeef3f9b2a609e6984f017a2f/libs/seiscomp/client/application.cpp#L273

gempa-jabe:
Why must they call it if it is the default? I can't follow your statement right now (it is Friday ;) ). Does this PR brake current behaviour with respect  to initialization?

damb:
Well, I'm not 100% sure, however, a call to e.g. `bindings2cfg --key-dir ./etc/key -o config.xml` previously succeeded without explicitly specifying a config DB. Now, it fails due to https://github.com/SeisComP/common/blob/df83362fbb6ea29eeef3f9b2a609e6984f017a2f/libs/seiscomp/client/application.cpp#L1050-L1053

if not setting the DB.

gempa-jabe:
Well then we have to fix bindings2cfg.

gempa-jabe:
Fixed with c127b41f8e1fc323148b257ff92afe555ae515e5.

damb:
Thanks.

gempa-jabe:
I have to reopen the PR. The fix with `setConfig` causes an application to crash if it is created after another application has been destroyed and both need to read the configuration. Actually we need to reset the instance and not the config object. I will fix that but check with you if you require `ConfigDB::setConfig(...)` or if it was only necessary for this particular issue.

gempa-jabe:
I have added a `Reset` method to all singletons which are called in the respective applications. That fixes a crash and furthermore also cleans up the loaded plugins.

damb:
@gempa-jabe, thanks for rechecking. The implemented solution works for me, too.

Originally, I used `ConfigDB::setConfig(...)` in order to have it aligned with `Inventory::setInventory(...)`. I wasn't aware of the fact that the underlying config object is reused under certain circumstances.

PR #53, opened 2022-03-02
[S-L2 picker]: allow to use L2 S picker on vertical componet only

luca-s:
I would like to propose this change that would allow the S-Picker L2 to work on the vertical component. My use case it that I have sensors with a single component only and some other sensors with 3 components, but whose orientation is not known. On both cases the current L2 picker cannot work because it requires 2 horizontal components to compute the L2-norm


luca-s:
My bad, I opened the PR on the wrong branch. I am closing this and opening a new one for the master branch

PR #54, opened 2022-03-02
[S-L2 picker]: allow to use L2 S picker on vertical component only

luca-s:
I would like to propose this change that would allow the S-Picker L2 to work on the vertical component. My use case it that I have sensors with a single component only and some other sensors with 3 components, but whose orientation is not known. On both cases the current L2 picker cannot work because it requires 2 horizontal components to compute the L2-norm

### DIFF
--- a/libs/seiscomp/processing/secondarypicker/S_v.cpp
+++ b/libs/seiscomp/processing/secondarypicker/S_v.cpp
@@ -18,14 +18,13 @@
  ***************************************************************************/
 
 
-#define SEISCOMP_COMPONENT L2Picker
+#define SEISCOMP_COMPONENT VPicker
 
 #include <seiscomp/logging/log.h>
 #include <seiscomp/processing/operator/ncomps.h>
-#include <seiscomp/processing/operator/l2norm.h>
 #include <seiscomp/math/filter.h>
 
-#include "S_l2.h"
+#include "S_v.h"
 
 
 using namespace std;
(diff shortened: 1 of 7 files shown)


luca-s:
This is what I thought too, but I decided to submit the PR with the minimum changes to see what was the idea of the other developers (actually you ;))

I will refactor the code a bit and push the change in here.

luca-s:
I was wandering if it is any good to keep the SL2Picker picker (and now the S-V picker) in the Seiscomp::Processing namespace. We could keep them in an anonymous namespace and register them without exposing them in the public API (which is what we recently  did we the `Balanced` and `Routing` RecordStreams).

luca-s:
Pushed the refactored code, please let me know if you spot more things to be cleaned up

libs/seiscomp/processing/secondarypicker/S_v.cpp
```diff
@@ -0,0 +1,108 @@
...
+	catch ( ... ) { cfg.detecFilter = "STALTA(1,10)"; }
+
+	return setAicConfig(cfg);
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+WaveformOperator* SVPicker::createFilterOperator(Filter* compFilter)
```
gempa-jabe, line 93: While thinking about it, this picker does not need an operator because it is just one component. `WaveformProcessor::setFilter` should be enough. We can keep it like that for the time being but it feels a bit overdesigned and introduces code that is actually not required.

luca-s: I agree and I thought thoroughly about it while creating the common base class for S_l2 and S_v. However this is the only way to make the common class work on both scenarios: it can handle multiple components but it is over designed for the basic scenario of a single one. I am happy to improve the code if you have some ideas.

gempa-jabe: Actually `setFilter` would need to receive a `ChainFilter` then. Of course, this could be implemented in case `createFilterOperator` returns `nullptr`. But I agree that this would not make the code more readable. We can keep it as it is now.

luca-s: IMO the issue is that S_l2 wants to filter the data before performing the L2-norm and only after it applies the detector filter. I wondert if that is really required. Why not applying the frequency filter and the detector both on the L2-norm?

luca-s: I understand the result would be different, but I am questioning if that would be worse or not.

gempa-jabe: Because the L2 norm is not symmetric, it is just positive. Furthermore we get rid of the offsets of each component when doing the filtering prior to combining.

PR #55, opened 2022-03-02
Mwp bugfix

jsaul:
The default radiation pattern was 1.0 but should have been 0.52 according to Tsuboi et al. (1995). This is fixed in this PR.


gempa-jabe:
Thank you.

PR #57, opened 2022-03-10
Iasp91 P S tables small fix

salichon:
Make this arrival tables more consistent
Avoid a negative value for the S arrival table


salichon:
Hi @gempa-jabe 
a rapid follow up on velocity table fix.
Happy to sign the `cla` 

cheers
jerome (GNS)

gempa-jabe:
Hi @salichon, I won't merge that easily. I need to confirm at least with @jsaul and even better, with tests. Those tables were added decades ago and I don't know the origin of those values.

jsaul:
I don't know the origin/reason of these offsets but t(0,0) != 0 cannot be correct. The offset does appear to affect 
other distances, too, but AFAICS "only" for zero depth.

This needs to be fixed. I will look into that in detail ASAP. Thanks to @salichon for pointing it out.

salichon:
Kia Ora and thanks @jsaul @gempa-jabe 
All good to me, 
i understand of course this needs some testing and so :) 
I made some quick check on my side although not as thoroughly or as lengthy as it could be. 
(without  the creation of these tables for instance).
This came out of a question about depth comparison for a shallow hypocenter with 2 methods but same iasp91 velocity model.
I wanted to exclude this negative S wave  value question from the discussion I had.
Cheers!

PR #58, opened 2022-03-30
[exporter] Fix broken unicode XML

andres-h:
Change ExportSink.write(unicode, length) to ExportSink.write(bytes) to fix unicode XML.

### DIFF
--- a/libs/swig/io.i
+++ b/libs/swig/io.i
@@ -94,6 +94,10 @@
   }
 }
 
+%typemap(directorin) (const char *data, int size) {
+    $input = PyBytes_FromStringAndSize($1, $2);
+}
+
 %import "math.i"
 %include "seiscomp/core.h"
 %include "seiscomp/io/database.h"
--- a/libs/seiscomp/io/exporter.cpp
+++ b/libs/seiscomp/io/exporter.cpp
@@ -59,25 +59,10 @@ struct SinkBuf : std::streambuf {
 		if ( pbase() == pptr() ) return 0;
 
 		int bytes = pptr() - pbase();
-		int keepbytes = 0;
-
-		// UTF-8 character can take up to 4 bytes.
-		// Avoid breaking the characters.
-		while (bytes > 0 && (pbase()[bytes - 1] & 0xc0) == 0x80 && keepbytes < 4) {
-			++keepbytes;
-			--bytes;
-		}
-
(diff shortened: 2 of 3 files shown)


libs/swig/io.i
```diff
@@ -94,6 +94,10 @@
   }
 }
 
+%typemap(directorin) (const char *data, int size) {
+    $input = PyBytes_FromStringAndSize($1, $2);
```
gempa-jabe, line 98: That means that we go from two parameters to simply one? That would break the API. The conversion from bytes and length seems to be appropriate although I don't know exactly what `PyBytes_FromStringAndSize` does. Is the length of the bytes objects the same as `size`?

Please note that (data, size) could also be  a binary blob, it must not be interpreted as string. I know that the current approach is also wrong but I am not sure if the fix solves all the issues.

andres-h: > That means that we go from two parameters to simply one? That would break the API.

Yes. Maybe we can somehow keep `size` (which is now equivalent to `len(data)` in Python, thus redundant), but changing the type of `data` from `str` to `bytes` breaks the (Python) API anyway.

> I don't know exactly what `PyBytes_FromStringAndSize` does. Is the length of the bytes objects the same as `size`?

Yes.

> Please note that (data, size) could also be a binary blob, it must not be interpreted as string.

This is exactly why we need `PyBytes_FromStringAndSize`. Previously, data was not only interpreted as string, but **unicode** string.

andres-h: Previously `len(data) != size`, because some unicode characters are longer than one byte.

gempa-jabe: I agree that this is not trivial. So actually keeping only the bytes array and omitting the size attribute is the best option.

gempa-jabe: Why does it actually not change `io.py` to reflect the new signature?

andres-h: I don't know. Apparently changing io.py is not needed because of variable arguments.

BTW, I changed wrappers manually, because I have a different SWIG version and it generates completely different wrappers. But I ran fdsnws tests and there were no errors.

gempa-jabe:
I merge it and generate the code again and add the typemap(in).

gempa-jabe:
I have pushed everything and update the Python Wrapper. There were other changes required as well.

PR #60, opened 2022-05-26
[slink]: `retries` option is not considered

luca-s:
The seedlink RecordStream interface doesn't not take into consideration the `retries` option.

### DIFF
--- a/libs/seiscomp/io/recordstream/slconnection.cpp
+++ b/libs/seiscomp/io/recordstream/slconnection.cpp
@@ -178,7 +178,7 @@ SLConnection::SLConnection()
 : RecordStream() {
 	_readingData = false;
 	_sock.setTimeout(300); // default
-	_maxRetries = -1; // default
+	_maxRetries = 0; // default
 	_retriesLeft = -1;
 	_useBatch = true;
 }
@@ -187,7 +187,7 @@ SLConnection::SLConnection(string serverloc)
 : RecordStream() {
 	_readingData = false;
 	_sock.setTimeout(300); // default
-	_maxRetries = -1; // default
+	_maxRetries = 0; // default
 	_retriesLeft = -1;
 	_useBatch = true;
 }
@@ -436,7 +436,6 @@ Record *SLConnection::next() {
 					SEISCOMP_DEBUG("Handshaking SeedLink server at %s", _serverloc.c_str());
 					handshake();
 					_readingData = true;
-					_retriesLeft = -1;
 				}
 				catch ( SocketTimeout &ex ) {
 					Core::msleep(500);


luca-s:
I don't know if this slips under the radar, but it should be pretty straightforward to merge. I have tested the PR and it is working fine.

gempa-jabe:
We use the retries option frequently and it is working. What exactly is that fix supposed to do?

libs/seiscomp/io/recordstream/slconnection.cpp
```diff
@@ -436,7 +436,6 @@ Record *SLConnection::next() {
 					SEISCOMP_DEBUG("Handshaking SeedLink server at %s", _serverloc.c_str());
 					handshake();
 					_readingData = true;
-					_retriesLeft = -1;
```
gempa-jabe, line 439: I think the intention was to break the loop if the handshake was successful (after X retries) but the connection failed during streams caused by either a network issue, a server restart or a timeout. If you configured a maximum amount of retries then it only affects the handshake, e.g. to wait until the server is available.

If streaming is aborted for some reason then we should repeat the handshake in any case and if it fails after X retries then it will cancel the recordstream.

gempa-jabe:
Maybe you can phrase your expectation of the behaviour under each condition (with and without retries configured):

1. What do you expect if the initial connect to the server fails (e.g. seedlink not running or no network connection)?
2. What do you expect if the streaming is aborted due to a server restart or a network failure?

libs/seiscomp/io/recordstream/slconnection.cpp
```diff
@@ -178,7 +178,7 @@ SLConnection::SLConnection()
 : RecordStream() {
 	_readingData = false;
 	_sock.setTimeout(300); // default
-	_maxRetries = -1; // default
+	_maxRetries = 0; // default
```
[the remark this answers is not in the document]
luca-s: I understand that, but I wanted to make the code consistent with the documentation: `retries - number of connection retry attempts, default: 0`

libs/seiscomp/io/recordstream/slconnection.cpp
luca-s: > if you configured a maximum amount of retries then it only affects the handshake, e.g. to wait until the server is available.

I was surprised by this behaviour and for this reason I debugged it, but If this is the intended behviour than I am happy with that. 

> If streaming is aborted for some reason then we should repeat the handshake in any case

This is exactly how it works after this PR.

> and if it fails after X retries then it will cancel the recordstream.

The total amount of failures is kept and it the sum of all the failuers that has happened either during the handshake or during the streaming

luca-s:
I am not against the current behaviour, that is retrying only during the handshake phase. I just didn't know it worked like that, so I debugged the behaviour and I thought It was a bug. We can close the issue then.

The only thing that is inconsistent is the documentation that states the default `retries` value is 0, but in the code is infinite.

gempa-jabe:
That is actually a good question, the code was written more than a decade ago. ;)

In general one could also implement it as you said. Just add every reconnect attempt (during handshake or streaming, does not matter) and break if the maximum number of retries is reached. I am actually fine with removing the line and have a clear definition of behaviour. Currently it seems rather vague.

luca-s:
Now that I discussed the issue with you I feel like someone wanted this exact behaviour, so I  feel like keeping the code as it is. However I let you take the final decision.

gempa-jabe:
I think the most important use case was to allow the seedlink connection act as an archive request with `timeout=1&retries=0`. I have never used `retries=12`.

PR #61, opened 2022-06-02
Comment code for future reference

luca-s:
@gempa-jabe can we add this comment for future reference so that it is clear the current behaviour is intended?


gempa-jabe:
That is actually not always the case. -1 means try indefinitely unless maxRetries has been set to a non-negative value.

PR #62, opened 2022-06-07
Update Flinn-Engdahl regions to match Wikipedia

jsaul:
Many of the Flinn-Engdahl region names were inconsistent, e.g. "Calif.-Baja Calif. Border Region" vs. "Near Coast of N. California" or "New Ireland Region, P.N.G." vs. "Near N. Coast of New Guinea, PNG." vs. "New Guinea, Papua New Guinea" etc.

This update makes them consistent with the region names as listed in Wikipedia, see https://en.wikipedia.org/wiki/Flinn%E2%80%93Engdahl_regions

For readability each entry also contains a comment containing the Flinn-Engdahl region number.


gempa-jabe:
Thanks a lot.

PR #63, opened 2022-06-14
Update ferdata.h

comoglu:
As explained here, Irian Jaya has renamed West Papua in 2007.

https://www.britannica.com/place/West-Papua


jsaul:
Yes, we have this on the radar and there are actually other regions that would have to be renamed, like "Queen Charlotte Islands" to "Haida Gwaii". The goal of the recent modifications was to first make the names consistent with the names listed on the [Wikipedia page](https://en.wikipedia.org/wiki/Flinn%E2%80%93Engdahl_regions). Ideally the names would be made configurable.

gempa-jabe:
According to the [Wikipedia entry](https://de.wikipedia.org/wiki/Papua_Barat) and the source you listed I am fine with merging that.

PR #65, opened 2022-07-08
[core]: add TTT for a homogeneous velocity model

luca-s:
Dear developers,

this is a proposal to add a new travel time table for a homogeneous velocity model. This is clearly a niche use case, but since I needed it I thought to share the code and make it available to other users. This is also useful for certain testing cases, for example when developing new modules or for generating unit testing code.

If you are interested in merging this PR you might be want to review the interface name, currently `homogeneous` and how the model name is currently used ("abused") to pass the constant P and S velocity to the model. This is clearly an abuse of the model name, but it is very convenient for the user.

Currently it looks like the following:

image

Luca

### DIFF
--- a/libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml
+++ b/libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<seiscomp>
+	<plugin name="homogeneous">
+		<extends>global</extends>
+		<description>Travel time table for a homogeneous velocity model</description>
+		<configuration>
+			<group name="ttt">
+			  <group name="homogeneous">
+				<description>
+				 Travel-time interface for homogeneous velocity models.
+				</description>
+					<group name="profile" link="ttt.homogeneous.tables">
+						<description>
+							Parameters defining the area where the profile applies and the P/S velocities.
+							Once defined, the profile can then be registered in ttt.homogeneous.tables
+						</description>
+						<struct type="homogeneous profile">
+							<description></description>
+							<parameter name="origin" type="list:double" unit="deg">
+								<description>Geographic origin of the region. Expects 2 values: latitude, longitude.</description>
+							</parameter>
+							<parameter name="radius" type="double" unit="km">
+								<description>Radius validity of the region.</description>
+							</parameter>
+							<parameter name="minDepth" type="double" unit="km">
+								<description>Min Depth validity of the region.</description>
+							</parameter>
+							<parameter name="maxDepth" type="double" unit="km">
+								<description>Max Depth validity of the region.</description>
+							</parameter> 
+							<parameter name="P-velocity" type="double" unit="km/s">
+								<description>P wave velocity.</description>
+							</parameter>
+							<parameter name="S-velocity" type="double" unit="km/s">
+								<description>S wave velocity.</description>
+							</parameter>
+						</struct>
+					</group>
+				</group>
+			</group>
+		</configuration>
+	</plugin>
+</seiscomp>
--- a/libs/seiscomp/seismology/ttt/homogeneous.cpp
+++ b/libs/seiscomp/seismology/ttt/homogeneous.cpp
@@ -1,0 +1,220 @@
+/***************************************************************************
+ * Copyright (C) gempa GmbH                                                *
+ * All rights reserved.                                                    *
+ * Contact: gempa GmbH (seiscomp-dev@gempa.de)                             *
+ *                                                                         *
+ * GNU Affero General Public License Usage                                 *
+ * This file may be used under the terms of the GNU Affero                 *
+ * Public License version 3.0 as published by the Free Software Foundation *
+ * and appearing in the file LICENSE included in the packaging of this     *
+ * file. Please review the following information to ensure the GNU Affero  *
+ * Public License version 3.0 requirements will be met:                    *
+ * https://www.gnu.org/licenses/agpl-3.0.html.                             *
+ *                                                                         *
+ * Other Usage                                                             *
+ * Alternatively, this file may be used in accordance with the terms and   *
+ * conditions contained in a signed written agreement between you and      *
+ * gempa GmbH.                                                             *
+ ***************************************************************************/
+
+#include <seiscomp/seismology/ttt.h>
+#include <seiscomp/core/strings.h>
+#include <seiscomp/math/geo.h>
+#include <seiscomp/datamodel/config.h>
+#include <seiscomp/system/environment.h>
+
+#include <string>
+#include <cmath>
+#include <vector>
+
+using namespace std;
+using namespace Seiscomp;
+
+namespace {
+
+/**
+ * Homogeneous
+ *
+ * A class to compute seismic travel times for an homogeneous (constant)
+ * velocity
+ */
+class Homogeneous : public TravelTimeTableInterface {
+	public:
+		/**
+		 * Construct a TravelTimeTable object for the specified model.
+		 */
+		Homogeneous() = default;
+		~Homogeneous() = default;
+
+		Homogeneous(const Homogeneous &other) = default;
+		Homogeneous &operator=(const Homogeneous &other) = default;
+
+		Homogeneous(Homogeneous &&other) = default;
+		Homogeneous &operator=(Homogeneous &&other) = default;
+
+	public:
+		bool setModel(const std::string &model) override;
+		const std::string &model() const override;
+
+		TravelTimeList *
+		compute(double lat1, double lon1, double dep1,
+		        double lat2, double lon2, double alt2 = 0.,
+		        int ellc = 1) override;
+
+		TravelTime
+		compute(const char *phase,
+		        double lat1, double lon1, double dep1,
+		        double lat2, double lon2, double alt2 = 0.,
+		        int ellc = 1) override;
+
+		TravelTime
+		computeFirst(double lat1, double lon1, double dep1,
+		             double lat2, double lon2, double alt2 = 0.,
+		             int ellc = 1) override;
+
+		bool isInside(double lat, double lon, double dep);
+
+	private:
+		std::string _model;
+		double _pVel = 1; // [km/s]
+		double _sVel = 1; // [km/s]
+		double _centerLat = 0; // [degree]
+		double _centerLon = 0; // [degree]
+		double _radius = 0; // [km]
+		double _minDepth = 0; // [km]
+		double _maxDepth = 0; // [km]
+};
+
+double radToDeg(double r) { return 180.0 * r / M_PI; }
+
+double degToRad(double d) { return M_PI * d / 180.0; }
+
+double computeDistance(double lat1, double lon1,
+                       double lat2, double lon2,
+                       double *azimuth = nullptr,
+                       double *backAzimuth = nullptr) {
+	double dist, az, baz;
+	Math::Geo::delazi(lat1, lon1, lat2, lon2, &dist, &az, &baz);
+	dist = Math::Geo::deg2km(dist);
+	if ( azimuth ) {
+		*azimuth = az;
+	}
+	if ( backAzimuth ) {
+		*backAzimuth = baz;
+	}
+	return dist;
+}
+
+bool Homogeneous::setModel(const string &model) {
+
+	// load global configuration
+	Config::Config cfg;
+	if ( ! Environment::Instance()->initConfig(&cfg, "") ) {
+		return false;
+	}
+
+	string base = "ttt.homogeneous.profile." + model + ".";
+	vector<string> origin;
+	try {
+		_pVel      = cfg.getDouble(base + "P-velocity");
+		_sVel      = cfg.getDouble(base + "S-velocity");
+		_radius    = cfg.getDouble(base + "radius");
+		_minDepth  = cfg.getDouble(base + "minDepth");
+		_maxDepth  = cfg.getDouble(base + "maxDepth");
+		origin = cfg.getStrings(base + "origin");
+	}
+	catch (...) {
+		return false;
+	}
+
+	if ( origin.size() != 2                         ||
+	   ! Core::fromString(_centerLat, origin.at(0)) ||
+	   ! Core::fromString(_centerLon, origin.at(1)) ) {
+		return false;
+	}
+
+	_model = model;
+	return true;
+}
+
+const string &Homogeneous::model() const {
+	return _model;
+}
+
+TravelTimeList *
+Homogeneous::compute(double lat1, double lon1, double dep1,
+                     double lat2, double lon2, double alt2, int ellc) {
+	TravelTimeList *ttlist = new TravelTimeList;
+	ttlist->delta = computeDistance(lat1, lon1, lat2, lon2);
+	ttlist->depth = dep1;
+	try {
+		ttlist->push_back(compute("P", lat1, lon1, dep1, lat2, lon2,  alt2, ellc));
+	}
+	catch (const NoPhaseError& e ) { }
+	try {
+		ttlist->push_back(compute("S", lat1, lon1, dep1, lat2, lon2,  alt2, ellc));
+	}
+	catch (const NoPhaseError& e ) { }
+	ttlist->sortByTime();
+	return ttlist;
+}
+
+bool
+Homogeneous::isInside(double lat, double lon, double dep)
+{
+	if ( dep < _minDepth || dep > _maxDepth ) {
+		return false;
+	}
+	double dist = computeDistance(lat, lon, _centerLat, _centerLon);
+	if ( dist > _radius ) {
+		return false;
+	}
+	return true;
+}
+
+TravelTime
+Homogeneous::compute(const char *phase,
+                     double lat1, double lon1, double dep1,
+                     double lat2, double lon2, double alt2, int ellc) {
+	double velocity;
+	if ( phase[0] == 'P' || phase[0] == 'p' ) {
+		velocity = _pVel;
+	}
+	else if ( phase[0] == 'S' || phase[0] == 's' ) {
+		velocity = _sVel;
+	}
+	else {
+		throw NoPhaseError();
+	}
+
+	if ( !isInside(lat1, lon1, dep1) ) {
+		throw NoPhaseError();
+	}
+
+	// straight ray path since we are in a homogeneous media
+	double Hdist = computeDistance(lat1, lon1, lat2, lon2);
+	double Vdist = dep1 + alt2/1000.;
+	double distance = sqrt(Hdist*Hdist + Vdist*Vdist); // [km]
+	double tt = distance / velocity; // [sec]
+	double takeOffAngle = atan2(Vdist, Hdist);
+	takeOffAngle += degToRad(90); // -90(down):+90(up) -> 0(down):180(up)
+	takeOffAngle = radToDeg(takeOffAngle);
+
+	return TravelTime(phase, tt, 0, 0, 0, takeOffAngle);
+}
+
+TravelTime
+Homogeneous::computeFirst(double lat1, double lon1, double dep1,
+                          double lat2, double lon2, double alt2, int ellc) {
+	if ( _pVel >= _sVel ) {
+		return compute("P", lat1, lon1, dep1, lat2, lon2,  alt2, ellc);
+	}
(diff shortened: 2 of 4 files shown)


libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <seiscomp/core/strings.h>
+#include <seiscomp/math/geo.h>
+#include <seiscomp/seismology/ttt/homogeneous.h>
+
+#include <cmath>
+#include <vector>
+
+using namespace std;
+
+namespace
```
gempa-jabe, line 29: Namespace and bracken in one line

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,129 @@
...
+                       double lon1,
+                       double lat2,
+                       double lon2,
+                       double *azimuth = nullptr,
+                       double *backAzimuth = nullptr)
+{
+	double dist, az, baz;
+	Seiscomp::Math::Geo::delazi(lat1, lon1, lat2, lon2, &dist, &az, &baz);
+	dist = Seiscomp::Math::Geo::deg2km(dist);
+	if (azimuth) *azimuth = az;
```
gempa-jabe, line 45: We use consequently spaces within parenthesis: `if ( azimuth )`

libs/seiscomp/seismology/ttt/homogeneous.h
```diff
@@ -0,0 +1,83 @@
...
+		        double lat2, double lon2, double alt2 = 0.,
+		        int ellc = 1) override;
+
+		TravelTime
+		computeFirst(double lat1, double lon1, double dep1,
+		             double lat2, double lon2, double alt2 = 0.,
+		             int ellc = 1) override;
+	private:
+		std::string _model;
+		double pVel; // [km/s]
```
gempa-jabe, line 74: As `pVel` is a private member please prefix it with an underscore.

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,129 @@
...
+double radToDeg(double r) { return 180.0 * r / M_PI; }
+double degToRad(double d) { return M_PI * d / 180.0; }
+
+double computeDistance(double lat1,
+                       double lon1,
+                       double lat2,
+                       double lon2,
+                       double *azimuth = nullptr,
+                       double *backAzimuth = nullptr)
+{
```
gempa-jabe, line 41: Please remove the newline before the curly bracket.

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+TravelTime
+Homogeneous::compute(const char *phase,
+                     double lat1, double lon1, double dep1,
+                     double lat2, double lon2, double alt2, int ellc)
+{
+	double velocity;
+	if ( phase[0] == 'P' || phase[0] == 'p') {
+		velocity = pVel;
+	} else if ( phase[0] == 'S' || phase[0] == 's') {
```
gempa-jabe, line 96: Add a newline before `else if`

libs/seiscomp/seismology/ttt/homogeneous.cpp
gempa-jabe: In general we are about to change the code towards curly brackets even for blocks with only one statement to avoid indentation errors. We ran several times into issues with that.

```cpp
if ( backAzimuth ) {
    *backAzimuth = baz;
}
```

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,129 @@
...
+namespace TTT {
+
+
+bool Homogeneous::setModel(const string &model)
+{
+	vector<string> tokens;
+	Core::split(tokens, model, "/", false);
+	if (tokens.size() != 2)
+		return false;
+	pVel = stod(tokens.at(0));
```
gempa-jabe, line 62: For string conversion I would prefer `Core::fromString` which also returns a success flag.

gempa-jabe (changes requested):

gempa-dirk:
@luca-s @gempa-jabe This is a useful extension. I recommend there should be depth (Moho/Konrad discontinuity) and distance limits if the implementation considers a flat but not a sphercial Earth. The limits could be even hardcoded.

gempa-dirk:
@luca-s @gempa-jabe Phases labelled p, Pg and s, Sg and maybe other phases should be considered as P and S, respectively.

gempa-jabe:
@luca-s are you aware that you can extend travel time table implementation with plugins? So this could easily go into a plugin. That makes things easier when it comes to incompatible changes to the API. A plugin is not affected by that.

luca-s:
@gempa-jabe Yes I am aware of that. I just thought to double check with you devs if this could be an useful extension for other users or not. I don't need to merge this, but If other users are interested in that I would be happy to contribute to seiscomp.

gempa-jabe:
Contributing to SeisComP is OK but maybe better as plugin. As this "debug" code is subject to many improvements and changes it would require incompatible changes and we need to increase the major version each time that happens. Another option is to not include its header in the SDK and to not expose any linker symbols (private namespace).

luca-s:
>  Another option is to not include its header in the SDK and to not expose any linker symbols (private namespace).

I like this very much

luca-s:
> Phases labelled p, Pg and s, Sg and maybe other phases should be considered as P and S, respectively.

That has been taken care of [here](https://github.com/luca-s/common/blob/1755a2879de881755b6a5dcbe1b9e5288fc76096/libs/seiscomp/seismology/ttt/homogeneous.cpp#L98) and [here](https://github.com/luca-s/common/blob/1755a2879de881755b6a5dcbe1b9e5288fc76096/libs/seiscomp/seismology/ttt/homogeneous.cpp#L101).

luca-s:
> I recommend there should be depth (Moho/Konrad discontinuity) and distance limits if the implementation considers a flat but not a sphercial Earth. The limits could be even hardcoded.

I don't mind adding those limits, however I wonder if that is really necessary. Since this model is too simple to be meaningful over a few km distance/depth, if we really want to constrain it we should focus on the range of 2,3 maybe 5 kms. However it is difficult to say how much it should extend for. Therefore I would let the model be unbound. If a user really makes use of this, they must also understand its limitations.

gempa-dirk:
@luca-s @gempa-jabe 

> If a user really makes use of this, they must also understand its limitations.

This assumption is very obvious but also very often violated. Make it configurable and use stringent constraints as defaults, e.g. 5 km.

luca-s:
I have refactored the code and now it resides in an anonymous namespace. I have also added the option to constrain maximum depth and maximum distance. However the configuration options are still passed via model name, which is ugly (i.e. model name is "vp/vs" or "vp/vs/depth/distance" - e.g. model name "5.3/3.2" or "5.3/3.2/10/20"). I believe I should extend the global configuration and add a group (e.g. `homogeneous` or `homogeneousTTT`) that contains the parameters required by this ttt. What do you think?

luca-s:
The configuration of the homogeneous TTT are now stored in `global.cfg` as multiple profiles. Please let me know if I need to modify anything.


Screenshot from 2022-08-02 12-32-58
Screenshot from 2022-08-02 12-33-21

gempa-dirk:
@luca-s It contains the relevant parameters but I have a few recommendations:

- call the parameter region.radius , not region.radious 
- instead of using 2 parameters centerLat and centerLon you cound use one, e.g. origin, just like for the nlloc plugin.
- velocity parameters should have a unit in the description XML
- I would drop the subsection "region": there is just one and a division only makes sense, if you have at least 2, the parameters become easier and shorter, the view in scconfig is less complex and shorter.

gempa-dirk:
@luca-s 

Important: The profiles of the Homogeneouns section must be a subsection of ttt.homogenous section, not a separate one. Then, you can also drop Homogeneous.

luca-s:
> Important: The profiles of the Homogeneouns section must be a subsection of ttt.homogenous section, not a separate one. 

This is not possible, see the [xml configuration](https://github.com/SeisComP/common/blob/dea079b3053ce0c76e57646bddcc924b4626ceb7/apps/descriptions/global.xml#L387) of `global.cfg`. However wE can add a subsection called `ttt.homogenous-profiles` containing the homogeneous parameters, but we still have to keep the `ttt.homogeneous` as it is now, that is a simple list of active profiles. What do you think?

luca-s:
I fixed the minor issues. The only remaining question if we have to move these paramters inside the `ttt` group  or leave them on their own.

Screenshot from 2022-08-02 16-13-44

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+		*backAzimuth = baz;
+	}
+	return dist;
+}
+
+bool Homogeneous::setModel(const string &model) {
+
+	// load global configuration
+	Config::Config cfg;
+	if ( ! Environment::Instance()->initConfig(&cfg, "") ) {
```
luca-s, line 112: @gempa-jabe Could you please double check this part of the code? It works but I have the feeling there are better ways to load the global configuration.

gempa-dirk:
@luca-s Without compiling your branch, would this not put all parameters in the same section:

`<?xml version="1.0" encoding="UTF-8"?>
<seiscomp>
	<plugin name="homogeneous">
		<extends>global</extends>
		<description>Travel time table for a homogeneous velocity model</description>
		<configuration>
			<group name="ttt">
				<group name="homogeneous">
					<description>
					Travel time table parameters for homogeneous velocity models, requires the plugin homogeneous.
					</description>
					<group name="profile">
						<description>
							Profiles containing the specific travel time table parameters.
						</description>
						<struct type="homogeneous profile" link = "ttt.homogeneous.tables">
							<description></description>
							<parameter name="origin" type="list:double" unit="deg">
								<description>Geographic origin of the region. Expects 2 values: latitude, longitude.</description>
							</parameter>
							<parameter name="radius" type="double" unit="km">
								<description>Radius validity of the region.</description>
							</parameter>
							<parameter name="minDepth" type="double" unit="km">
								<description>Min Depth validity of the region.</description>
							</parameter>
							<parameter name="maxDepth" type="double" unit="km">
								<description>Max Depth validity of the region.</description>
							</parameter> 
							<parameter name="P-velocity" type="double" unit="km/s">
								<description>P wave velocity.</description>
							</parameter>
							<parameter name="S-velocity" type="double" unit="km/s">
								<description>S wave velocity.</description>
							</parameter>
						</struct>
					</group>
				</group>
			</group>
		</configuration>
	</plugin>
</seiscomp>`

luca-s:
@gempa-dirk Yes it works, but I wasn't clear with my previous comment where I tried to explain the issues I see with that.  The thing I don't like is that `homogeneous` appear twice: one instance is used to configure the profiles and the second time is used to specify the list of available profiles for the GUIs.

image

gempa-dirk:
@luca-s We currently have the same issues with other global configurations such as regionalized magnitudes where we decided to go for the same solution. So I recommend to use it here, too. Maybe in the future we find an optimized way for arranging in scconfig but definitely the parameters belong together. Or practically spoken: If you configure without scconfig and in the config files the grouping is then consistent as:

```
ttt.homogeneous.tables =
ttt.homogeneous.profile.$name.[param] = 
```

Otherwise it becomes very complicated.

luca-s:
@gempa-dirk thanks, that makes sense. I'll push the change then

luca-s:
Done. I have also quashed all the commits in one.

luca-s:
I have used this code for some time and it seems to work well. If you are happy with structure of the configuration and if @gempa-jabe confirms the code is ok, then this PR is completed.

gempa-dirk:
@luca-s The XML is o.k. now for the profiles. I would add a note to the descruption under "homogeneous" that profiles must be defined and registered in ttt.homogeneous.tables

Have you added a description for ttt.homogeneous.tables   ?

luca-s:
@gempa-dirk The configuration looks like the following now:

image


I didn't add a description file (.rst) because there is currently no description for the `ttt` parameters and since this is a  sub-configuration of `ttt` there is no right place to put a .rst file. Also I believe the current description in global.xml is quite self-explanatory. What do you think?

gempa-dirk:
> @gempa-dirk The configuration looks like the following now:
> 
> image
> 
> I didn't add a description file (.rst) because there is currently no description for the `ttt` parameters and since this is a sub-configuration of `ttt` there is no right place to put a .rst file. Also I believe the current description in global.xml is quite self-explanatory. What do you think?

**Configuration looks good now.**

If you do not want to write an rst you could have all information in the XML. I would re-write the first sentence of the profile description and move it in parts one level up below "homogenous" avoiding too much repetition with multiple profiles, e.g. "Travel-time interface for homogeneous velocity models". And below profiles you may shorten to the last 2 sentences: "Parameters defining the area ..."

We may eventually need to add "homogeneous" to the built-in interfaces listed in the description below "ttt".

luca-s:
Like this?

image

gempa-dirk:
Would be perfect for me.

luca-s:
I pushed the commits.

luca-s:
> I pushed the commits.

actually no, let me do it now

gempa-dirk:
O.k., then let's wait for @gempa-jabe to review.

libs/seiscomp/seismology/ttt/homogeneous.cpp
gempa-jabe: If an application instance is available, you can use

```c++
SCCoreApp->configGetDouble(base + "P-velocity");
```

If not or if you also want that code to be used without an application instance then this is the right way to go. It is probably not the most performant way to do it as the configuration is static and just needs to be read the first time.

libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml
```diff
@@ -0,0 +1,43 @@
...
+			  <group name="homogeneous">
+				<description>
+				 Travel-time interface for homogeneous velocity models.
+				</description>
+					<group name="profile">
+						<description>
+							Parameters defining the area where the profile applies and the P/S velocities.
+							Once defined, the profile must be registered in ttt.homogeneous.tables
+						</description>
+						<struct type="homogeneous profile" link = "ttt.homogeneous.tables">
```
gempa-jabe, line 17: Actually the "link" is not present in this description. As far as I see the code is not relying on traversing available tables so the registration is not necessary.

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+TravelTimeList *
+Homogeneous::compute(double lat1, double lon1, double dep1,
+                     double lat2, double lon2, double alt2, int ellc) {
+	TravelTimeList *ttlist = new TravelTimeList;
+	ttlist->delta = computeDistance(lat1, lon1, lat2, lon2);
+	ttlist->depth = dep1;
+	try {
+		ttlist->push_back(compute("P", lat1, lon1, dep1, lat2, lon2,  alt2, ellc));
+	} catch (const NoPhaseError& e ) { }
```
gempa-jabe, line 151: Please add a line break before `catch`.

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+	try {
+		ttlist->push_back(compute("S", lat1, lon1, dep1, lat2, lon2,  alt2, ellc));
+	} catch (const NoPhaseError& e ) { }
```
gempa-jabe, line 154: Same here: line break

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+TravelTime
+Homogeneous::compute(const char *phase,
+                     double lat1, double lon1, double dep1,
+                     double lat2, double lon2, double alt2, int ellc) {
+	double velocity;
+	if ( phase[0] == 'P' || phase[0] == 'p' ) {
+		velocity = _pVel;
+
+	} else if ( phase[0] == 'S' || phase[0] == 's' ) {
```
gempa-jabe, line 180: Please add a line break before `else`.

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+		velocity = _sVel;
+
+	} else {
```
gempa-jabe, line 183: Line break ...

libs/seiscomp/seismology/ttt/homogeneous.cpp
```diff
@@ -0,0 +1,217 @@
...
+	return TravelTime(phase, tt, 0, 0, 0, takeOffAngle);
+}
+
+TravelTime
+Homogeneous::computeFirst(double lat1, double lon1, double dep1,
+                          double lat2, double lon2, double alt2, int ellc) {
+	if ( _pVel >= _sVel ) {
+		return compute("P", lat1, lon1, dep1, lat2, lon2,  alt2, ellc);
+
+	} else {
```
gempa-jabe, line 209: Line break ...

gempa-jabe (changes requested):

gempa-jabe:
The code looks pretty nice and clean and we can merge it after resolving the remaining tiny issues. I haven't tested the functionality though.

libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml
luca-s: I'll move the "link" in the description of group "profile", where the link is actually used. I'll also change "must be registered" with "can then be registered"

luca-s:
I pushed the changes.

luca-s:
>  I haven't tested the functionality though.

I have been using this code for some time now, with scolv and scanloc. It woks fine as far as I can tell.

gempa-jabe (approved):

gempa-jabe:
Thank you. Then let's merge it!

luca-s:
Thank you @gempa-dirk @gempa-jabe

luca-s:
@gempa-dirk could you double check that you have pulled the latest changes from github? [libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml](https://github.com/SeisComP/common/blob/master/libs/seiscomp/seismology/ttt/descriptions/global_homogeneous.xml) is in the master branch now.

gempa-dirk:
Yes, and in your XML global_homogeneous.xml you are have no ttt.homogeneous.tables, like:

`<parameter name="ttt.homogeneous.tables" type="string" unit="">
								<description></description>
							</parameter>
`

You just have the link and you probably have the parameter defined in global.xml or soimewhere else.

luca-s:
Currently the configuration looks like this:

Screenshot from 2022-08-26 09-54-25

...but you would like to see `ttt.homogeneous.tables`  visible by default below `LOCAST` and `libtau`? 

I can provide a PR for that.

luca-s:
PR #69

gempa-jabe:
Actually `ttt.homogeneous.tables` is not being used by this implementation and therefore is not necessary to be provided. We only use those table definitions in the GUI to be able to populate the dropdown box. But this is not part of the TTT implementation and a decision of the GUI code which parameter to read.

luca-s:
I'll let you decide. However providing an empty list for `ttt.homogeneous.tables` makes more obvious to the user that this model can be added to the GUI if they want

luca-s:
@gempa-jabe by the way, the #69 simply adds an empty value for `ttt.homogeneous.tables` in `global.cfg`. I didn't mean to mess around with the xml

gempa-jabe:
Yes, that is correct. I also replied to @gempa-dirk. Providing the empty list is totally fine with me.

gempa-dirk:
@gempa-jabe @luca-s I think I have triggered this confusion. A simple solution seems to be to add a statement in the description of the of  ttt.homogeneous.profile that the use must add another ttt profile named "homogeneous" and add the name of the profile to the tables parameter.

Then drop the empty parameter ttt.homogeneous.tables from global.cfg

BTW: I havetested the interface with local events and worked for me.

Issue #66, opened 2022-07-08
[GUI] picker/amplitude views show wrong theoretical arrivals when origin has negative depth


luca-s:
The theoretical arrivals on the picker and amplitude GUIs are not properly computed when an origin has a negative depth. This is because the negative depth is set to +1km when calling the travel time table implementations . See [here](https://github.com/SeisComP/common/blob/4a85104ea1fe99cc8ce1ef012909d20ad3d2f8a5/libs/seiscomp/gui/datamodel/amplitudeview.cpp#L4274) and [here](https://github.com/SeisComP/common/blob/4a85104ea1fe99cc8ce1ef012909d20ad3d2f8a5/libs/seiscomp/gui/datamodel/pickerview.cpp#L4899).

While I understand the "safety belt" intentions behind this choice, it would be much better to add the depth check in the travel time table implementation themselves. That would allow other TTT implementations that handle negative depths to work properly.

What do you think?

Thanks,
Luca

jsaul:
I don't remember the rationale of this 1-km setting. It may be related 
to some numerical issues related to exact-zero depth, which however 
could be fixed by setting the depth to just 1 meter (not kilometer). 
That was long ago.

But I agree with what you say. The current tables are not correct for 
extremely shallow or negative depths and therefore need to be fixed.

luca-s:
So my idea would be to remove the two lines mentioned above and add some depth constraints in the travel time table implementations. `libtau` has [already a check](https://github.com/SeisComP/common/blob/968d7d13c0ed3d51311159dcafdc5da06ed26ee4/libs/seiscomp/seismology/ttt/libtau.cpp#L133) for that , so I could add a check to LOCSAT ttt too.

gempa-jabe:
Yes, that makes much more sense to me.

PR #67, opened 2022-07-13
[GUI] allow theoretical arrivals with negative depth in picker/amplitude guis

luca-s:
This is related to issue #66 

The code forces same depth of libtau 0.01 km. That is for consistency

### DIFF
--- a/libs/seiscomp/gui/datamodel/amplitudeview.cpp
+++ b/libs/seiscomp/gui/datamodel/amplitudeview.cpp
@@ -4271,8 +4271,6 @@ bool AmplitudeView::addTheoreticalArrivals(RecordViewItem *item,
 			depth = 0.0;
 		}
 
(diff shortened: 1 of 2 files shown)


gempa-jabe:
The question is if we want to clamp depth values to 0.1 or throw an NoPhase exception if LOCSAT does not return phases. Otherwise a user could think that for depth -3km the travel times are correct. I would forward the depth to the 3rd party function and return its results. LOCSAT should be pretty stable in interpolating the grids so negative depths should be OK. But I have to admit that I haven't tested it yet.

luca-s:
> I would forward the depth to the 3rd party function and return its results.

I like that. I have tested LOCSAT ttt with negative depths and it seems to work. I just don't know how reliable the values are, but this seems a better way of handling this PR

PR #68, opened 2022-07-26
Feature/update flinn engdahl region names

nackerley:
The Queen Charlotte Islands were renamed Haida Gwaii in 2010. See https://en.wikipedia.org/wiki/Haida_Gwaii#Naming. This should be reflected in the names of the Flinn Engdahl regions, out of respect for the Haida people.

This issue is mentioned in https://github.com/SeisComP/common/pull/63, but not fixed in the associated pull request.

I have updated both names.asc and the ferdata.h, including the name change of Irian Jaya to West Papua described in https://github.com/SeisComP/common/pull/63. Although I suspect only the latter is used by SeisComP, the former could be useful to other organizations.

### DIFF
--- a/libs/seiscomp/seismology/regions/fedata/names.asc
+++ b/libs/seiscomp/seismology/regions/fedata/names.asc
@@ -19,7 +19,7 @@ Southern Yukon Territory, Canada
 Southeastern Alaska
 Off Coast of Southeastern Alaska
 West of Vancouver Island
-Queen Charlotte Islands Region
+Haida Gwaii Region
 British Columbia, Canada
 Alberta, Canada
 Vancouver Island, Canada Region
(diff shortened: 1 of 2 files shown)


jsaul:
As mentioned in https://github.com/SeisComP/common/pull/63 the reason for initially not renaming Queen Charlotte Islands to Haida Gwaii was solely to first make the names in SC consistent with those of Wikipedia, in spite of the undisputed fact that Haida Gwaii is more appropriate.

nackerley:
> ... the reason for initially not renaming Queen Charlotte Islands to Haida Gwaii was solely to first make the names in SC consistent with those of Wikipedia ...

Done: https://en.wikipedia.org/wiki/Flinn%E2%80%93Engdahl_regions#Southeastern_Alaska_to_Washington.

Several of us at NRCan are simultaneously working on updating the source files at the USGS, in particular at https://earthquake.usgs.gov/data/flinn_engdahl.php. This is a very important issue; we simply must start fixing it, now. We can't have everyone waiting for somebody else to fix it, first.

I expect that there will be other regions that need renaming, too. Let's just deal with them, one at a time, whenever there is clear consensus.

gempa-jabe:
@cla-bot check

gempa-jabe:
@cla-bot check

jsaul:
@nackerley
Thanks a lot for also making the change in Wikipedia!

PR #69, opened 2022-08-26
[core] visualize ttt.homogeneous.tables in scconfig by default

luca-s:
image


gempa-dirk:
@luca-s maybe the default should be no value instead of empty string, so no '""'

PR #71, opened 2022-09-05
[picker] Included in pick tooltip: publicID, time

jsaul:
Added a few useful informations to the pick tooltip: publicID, time, comments (if there are any)

### DIFF
--- a/libs/seiscomp/gui/datamodel/pickerview.cpp
+++ b/libs/seiscomp/gui/datamodel/pickerview.cpp
@@ -913,6 +913,12 @@ class PickerMarker : public RecordMarker {
 			}
 			catch ( ... ) {}
 
+			if ( !_referencedPick->publicID().empty() )
+				text += QString("\npublic ID: %1").arg(_referencedPick->publicID().c_str());
+			try {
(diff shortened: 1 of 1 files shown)


jsaul:
An issue related to the comments is that the comments are only shown for picks arriving via messaging, whereas when loaded from database the pick comments are _not_ included. Not sure if it would a performance issue to load the comments or if loading pick comments should be made optional (kind of like "load all picks").

gempa-jabe:
Loading pick comments has to be added. I think that the performance penalty is not too much if retrieved with one query rather than hundreds.

jsaul:
In fact I tried loading the comments already and it is hard to notice a difference.

jsaul:
Added config setting "loadPickComments" but need to actually propagate the setting to where the picks are loaded. Though maybe with a dedicated query it is not even needed to make this optional?

jsaul:
Forget about the comments, I don't need to display them there any more. Pick ID and time is nevertheless useful. AFAICS can be merged.

jsaul:
Let me know if there is anything that's blocking the merge.

gempa-jabe (approved):

PR #72, opened 2022-10-11
[xml] Always include preferred magnitude in quakeML generated with XSLT

acarapetis:
When an event's preferredMagnitude is contained in an origin that is not referenced by the event, this magnitude is omitted from QuakeML output.  For example, this is the case when dealing with scml produced by `scxmldump -fpFMAP` for an event with a moment magnitude preferred.

There's a note in the code that "every node not referenced by an event will be lost during the conversion", but I think this case is particularly counterintuitive because the magnitude *is* referenced by the event. Since magnitudes are direct children of events in QuakeML, I think this change makes sense.

### DIFF
--- a/libs/xml/0.10/sc3ml_0.10__quakeml_1.2.xsl
+++ b/libs/xml/0.10/sc3ml_0.10__quakeml_1.2.xsl
@@ -176,6 +176,9 @@
  *    - Skip originUncertaintyDescription if value is set to
  *      'probability density function' not supported by QuakeML.
  *
+ *  * 11.10.2022: Always output preferred magnitude (even when containing
+ *                origin is not referenced by event)
+ *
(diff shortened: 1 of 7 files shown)


gempa-stephan:
This pull request was overlooked. Sorry for keeping you waiting for so long. We tested your modifications and are about to merge them. Could you please resolve the merge-conflicts in the change log entries.

gempa-jabe:
Is the PR still of interest or can we close it without merging? There are still conflicts which must be resolved in order to merge the changes.

acarapetis:
@gempa-jabe Sorry, I missed Stephan's earlier comment. I will try to resolve the conflicts soon.

gempa-jabe:
@acarapetis ping. If you need help then we can check it.

PR #74, opened 2022-10-31
[xml] Improve performance in scml to quakeml XSLT

acarapetis:
SCML documents describing large earthquakes were taking a long time to convert to QuakeML, with almost all that time being spent in the loop that outputs picks that aren't referenced by amplitudes. (For example, an event with ~600 arrivals in its preferred origin was taking 12 seconds.)

By precomputing the list of referenced amplitudes once per origin (rather than computing it on the fly for each pick), performance is greatly improved (down to 400ms in my 600-arrival example).

I've got another PR open touching the same files (#72), but I figured I'd keep this separate because it doesn't change the output.

### DIFF
--- a/libs/xml/0.10/sc3ml_0.10__quakeml_1.2.xsl
+++ b/libs/xml/0.10/sc3ml_0.10__quakeml_1.2.xsl
@@ -176,6 +176,8 @@
  *    - Skip originUncertaintyDescription if value is set to
  *      'probability density function' not supported by QuakeML.
  *
+ *  * 31.10.2022: Improve performance when processing origins with many arrivals.
+ *
  ********************************************************************** -->
 <xsl:stylesheet version="1.0"
         xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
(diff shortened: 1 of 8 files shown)


gempa-stephan:
Thanks for your contribution. Using a variable here makes sense.

Would you mind sharing your test XML file as I couldn't reproduce the numbers. I used a file with ~400 amplitudes, 250 picks, 250 arrivals and converted it within 20ms, both, with the current and your improved XSLT script. 

Also let me know which XSLT processor you used. It was xsltproc in my case.

acarapetis:
Hi Stephan,

I'm also using xsltproc:
```
$ xsltproc -V
Using libxml 20910, libxslt 10134 and libexslt 820
xsltproc was compiled against libxml 20910, libxslt 10134 and libexslt 820
libxslt 10134 was compiled against libxml 20910
libexslt 820 was compiled against libxml 20910
```

First noticed the slowness when loading the file into obspy (which uses lxml + seiscomp's xslt to convert to quakeml).

Here's the XML I'm working with. There are actually less than 300 picks, but there are three origins and eight magnitudes referencing them, which might be causing a small combinatorial explosion?

[ga2022rlubky.xml.txt](https://github.com/SeisComP/common/files/9924946/ga2022rlubky.xml.txt)

gempa-stephan:
Sorry for the long delay and thanks for the XML example. I could now see the speedup: 0.216s compared to previous 5.227s.

Thanks!

PR #76, opened 2023-01-10
Patch of getThreeComponents()

jsaul:
In the special case of the component dip being -90° or +90° the component azimuth is undefined, unneeded and sometimes not set. This patch prevents such a stream from being ignored or processed incorrectly if the azimuth is not set.


gempa-jabe:
Thank, I am going to reformat the changes later.

PR #77, opened 2023-01-21
[butterworth filter] fix check on nyquist frequency

luca-s:
I have some data with a sampling rate of 2000hz and I used a butterworth filter of 100~1000hz. The code complained about the 1000hz (`frequency cutoff must be lower than Nyquist frequency`). I might be wrong, but shouldn't 1000hz be valid?

### DIFF
--- a/libs/seiscomp/math/filter/butterworth.cpp
+++ b/libs/seiscomp/math/filter/butterworth.cpp
@@ -354,7 +354,7 @@ void init_bw_biquads_inplace(Biquads &biquads, size_t order, double fmin, double
 			if ( fmax <= 0.0 )
 				throw std::runtime_error("High frequency cutoff must be greater than zero");
 
-			if ( fmax >= fnyquist )
+			if ( fmax > fnyquist )
 				throw std::runtime_error("High frequency cutoff must be lower than Nyquist frequency");
 
 			break;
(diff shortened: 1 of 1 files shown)


gempa-jabe:
In theory it should be valid. In practice it does not work correctly. Have your verified the filtering of 10-999 Hz and 10-1000 Hz with your patch? I have tested that for some data and the closer you get to the Nyquist frequency the less effective is the filter. For example filtering 200 sps data from 10-100 Hz does (visually) nothing whereas 10-99.999 Hz made a difference. Is a high-pass filter with 100 Hz not a better option?

jsaul:
AFAICS it does work correctly. The recursive filter simply cannot filter beyond Nyquist and in practice also has issues with cutoff frequencies close to but below Nyquist. I don't remember the gory details but this is somehow related to the bilinear transform used to map the analog to a digital filter. The bilinear transform has limitations and the closer the cutoff frequency gets to Nyquist the more the transition band will be "squeezed" between the cutoff and Nyquist.

With an analog filter in the frequency band you don't have these issues. Also, a highpass should have no issues with fc up to Nyquist.

luca-s:
I thought it was a [off-by-one error](https://en.wikipedia.org/wiki/Off-by-one_error) on the check, but it seems there are reasons for excluding to the Nyquist. That is fine with me. However the users will be surprised as me by the error message as they know the Nyquist freq should be valid.

jsaul:
For a bandpass or lowpass, I am pretty sure that choosing the Nyquist frequency as upper cutoff frequency is not valid. It might even be appropriate to issue a warning if the upper cutoff frequency is too close to Nyquist as distortion of the response function becomes inevitable.

BTW, are you sure that you want to bandpass filter if the upper cutoff frequency shall be at Nyquist? In that case, what you really want is a highpass, I believe.

luca-s:
You might be right that I should use a highpass filter. I wasn't really thinking thoroughly at the matter because I was told what configuration to use and I applied it and noticed the error.

PR #79, opened 2023-02-07
[sdsarchive] crash if used within a concurrent RS

luca-s:
The SDSArchive::close() leaves the class in an inconsistent class state and the next call to SDSArchive::next() crashes.

I noticed this behaviour when using `sdsarchive` within a `routing` record stream. The debugger showed the following:

image


gempa-jabe:
OK, the behaviour is different from `std::vector`. `std::list<T>::iterator() != std::list<T>().end()` while `std::vector<T>::iterator() == std::vector<T>().end()`. Thanks for catching that.

PR #82, opened 2023-02-15
 [io] Fix deadlock in Concurrent recordstream

luca-s:
When an acquisition thread terminates ( `Concurrent::acquiThread`  retuns) there is nobody closing the record queue and the main thread loops forever

### DIFF
--- a/libs/seiscomp/io/recordstream/concurrent.cpp
+++ b/libs/seiscomp/io/recordstream/concurrent.cpp
@@ -244,6 +244,11 @@ void Concurrent::acquiThread(RecordStream *rs) {
 	}
 
 	SEISCOMP_DEBUG("Finished acquisition thread");
+
+	if ( --_nthreads == 0) {
+		SEISCOMP_DEBUG("Closing record queue");
+		_queue.close();
+	}
 }
 // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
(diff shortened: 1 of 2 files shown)


gempa-jabe:
Thank you, that makes perfectly sense. I am going to test that soon.

luca-s:
@gempa-jabe please wait before merging because I noticed that the fix creates another issue somewhere else.

luca-s:
The problem is that I cannot acquire the mutex in `Concurrent::acquiThread` otherwise there is a deadlock if `Concurrent::close()` is called (precisely when `thread.join()` is called, since the thread is waiting to acquire the lock).

luca-s:
Now the original bug is fixed without introducing new bugs

gempa-jabe:
Thank you @luca-s. Using `atomic` seems like a overkill to me and indicates some synchronization issues. I will look into it and try to understand it.

gempa-jabe:
I have added a fix to this repo. Please check it with your use cases.

Note that using `atomic` breaks binary compatibility and a fix would not be possible for versions below 6.x.

luca-s:
You fix makes sense. I will test the patch with my scenarios that show the issue and I will report back if i still see a problem.

luca-s:
I confirm the problem is fixed, thanks.

jsaul:
Let me add that this issue seems to occur only with 'balanced'. A 'slink' RecordStream works fine even with an empty request. Hence the possible relation with this PR.

andres-h:
Is that the case still with the current version?

jsaul:
Yes, it is the case with the current version, _although_ I would not say "still" as this may well be unrelated to the already fixed issue.

As a workaround it is of course easy to prevent empty requests, but there may well be code that doesn't check for empty requests explicitly.

gempa-jabe:
This issue affects all recordstreams which inherit from `Concurrent`. 77f269ee07a947f2a37e51685c44aed50b66c81a should fix the issue.

Issue #83, opened 2023-02-27
Proposal to add a local seismicity locator


luca-s:
Dear developers,

I have been working on a couple of projects where they need a locator for very local seismicity (few kms and few hundreds of meters).  In this scenario I cannot use the default SeisComP locators because they do not take into consideration station elevations, negative source depths  and neither their  travel time table models do.

 We also use borehole sensors , which means the locator has to take into consideration that the earthquakes can happen also above the sensors. Basically many assumptions valid for regional seismicity break at this small scales. 

The only locator that works pretty well is NonLinLoc, but to achieve a good location resolution the number of iterations for the Octtree method has to be set  so high that it doesn't fit well for real-time applications, especially if you intend to use NonLinLoc within scanloc.

So I wrote [this simple locator ](https://github.com/swiss-seismological-service/genloc), which is the perfect complement to the `homogeneous` travel time table.

I wonder if you are interested in merging this code into SeisComP, since I don't get anything from keeping it in a separate repository. In case you are interested, just let me know and I will provide a PR where we can discuss improvements and changes required for the integration.

best
Luca

gempa-jabe:
That sounds great @luca-s. Not that I have tested your locator but I am pretty confident that something that complements your homogeneous travel time tables and serves as an additional locator options is a great and valuable addition. And the fact that your are still active with development so that bugs and issues are going to be fixed makes this integration very tempting. I have no issues accepting a PR. I only have to check formatting of the source ;).

Btw, as more algorithms require standard code such as fitting methods or statistical methods it would be great to put them all into one place and homogenize them. I know that this is not gonna be fun but the `3rd-party` directories are kind of hidden and other code won't be able to re-use it, especially in a plugin.

luca-s:
Good, then I will provide the PR where we can discuss changes in more details. I am pretty sure there are user level configuration that you might want to modify, clarify and what not.  

Should I follow the homogeneous approach where I put GenLoc code in a anonymous namespace to avoid polluting the SeisComP one?

luca-s:
We can move the discussion here #84

PR #84, opened 2023-03-06
[stdloc] add locator with focus on local seismicity

luca-s:
As discussed #83 this PR provides the inclusion of a new locator.  You can have a look at the current implementation and suggest any improvements or changes you would like to have before the integration. 

I have tested the code with local seismicity, but also at regional scale. I can see the results are pretty consistent with LOCSAT. I didn't write the locator with regional seismicity in mind, but the method is general enough so it works anyway and it makes testing easier.

One feature that I would like to add is the computation of the uncertainty ellipsoid. Hopefully I will find the time to add it soon.


plugins/locator/genloc/genloc.cpp
```diff
@@ -0,0 +1,1331 @@
...
+using IDList = Seiscomp::Seismology::LocatorInterface::IDList;
+using LocatorException = Seiscomp::Seismology::LocatorException;
+using StationNotFoundException = Seiscomp::Seismology::StationNotFoundException;
+using PickNotFoundException = Seiscomp::Seismology::PickNotFoundException;
+
+namespace {
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+ADD_SC_PLUGIN("Generic Locator", "gempa GmbH <seiscomp-devel@gempa.de>", 1, 0,
```
gempa-jabe, line 57: Please add yourself as author here.

gempa-jabe:
Just to avoid misunderstandings, we are not talking about a port of https://www.sciencedirect.com/science/article/pii/S0098300404001438 to SeisComP, right? Both projects just share the name by coincidence?

Otherwise it might be helpful to include a reference into the documentation or add a note that it is not to be confused with GENLOC.

luca-s:
> Just to avoid misunderstandings, we are not talking about a port of https://www.sciencedirect.com/science/article/pii/S0098300404001438 to SeisComP, right? Both projects just share the name by coincidence?
 
Just a coincidence.

luca-s:
@gempa-jabe I am now reading the article you linked. The similarities are impressive and I am tempted to change name

gempa-jabe:
Maybe we are talking about the similar method published in another paper unless you came up with that locator in a dark room by yourself. ;)

The GENLOC method is implemented in a commercial module (not ours) and maybe that could cause conflicts. So renaming it might be wise.

luca-s:
Well, the methods are basic and well known, that's why we both come up with the name gen(eric)loc. However I used, as a reference, the book "Routine Data Processing in Earthquake Seismology". 

I should have checked the name was not already in use. I will rename this locator.

luca-s:
For now let's call it `stdloc`: standard method locator. I open to better names

gempa-jabe:
`stdloc` sound good to me and avoids ambiguities with other locators.

gempa-jabe:
Do you think that the PR is ready to be merged? We can continue at any time in the future improving it.

luca-s:
I have tested it and I am happy with the results,  so you can integrate it if you like.  I still would like to add the confidence ellipsoid to the location, since at the moment there is no uncertainty reported. However, as you said, I can add it in a future PR. 

I don't know if there is someone planning to give it a try before publishing the code in  a SeisComP release. I would be happy if someone could give a critical review at the configuration parameters. Just to be sure it is clear the way they are presented and documented.

luca-s:
Sounds good, thanks.

gempa-jabe:
It seems your tab key has been stolen :wink:

luca-s:
Lol, It's just that I run the code though clang-format with default settings.

PR #87, opened 2023-03-18
[stdloc] better solutions for not well constrained events

luca-s:
Change some default settings to allow better locations  on events with small amount of picks or with outliers.  That is important when using stdloc in combination with scanloc for example.

This is a minor change for well constrained locations.

Also lower level of log messages, since they might be useful only when debugging,but they are annoying when using stdloc in normal operations.


gempa-jabe:
I have reformatted the code according to our existing style and will pull your changes directory into it. There are just a few.

luca-s:
all good, thank you

PR #89, opened 2023-03-30
[stdloc] bug fix + enhancement

luca-s:
This PR contain a bug fix, thanks to @FMassin that noticed there was an error in the computation of the arrival travel time residuals.

Also I added an option to control how the pick weights are computed from the time uncertainties. That was previously hardcoded.


gempa-jabe:
Enforcing our coding style does not seem to make much sense ;)

PR #90, opened 2023-03-31
Compute mb magnitude according to the IASPEI Magnitude WG recommendations

jsaul:
This PR implements the computation of mb according to the IASPEI Magnitude WG recommendations. There are some differences in the technique for determining the relevant extrema in the waveforms but the most notable difference is the correction by the WWSSN-SP amplitude response. This correction was not done previously and will have some effect especially for events for which the WWSSN-SP filtered P waveforms have a dominant period of more than one second. We will then observe on average a slight mb _increase_. For shorter periods mb will slightly _decrease_.

This is currently off by default and can be enabled by setting `amplitudes.iaspei = true` in the config. I am open to enable it by default, though.

Thanks to Jens Havskov for triggering this!

Note that this PR also implements a not directly related enhancement, namely the inclusion of the scaling time. At the moment only for mb and mB. The scaling time is the time at which the measurement (usually the amplitude maximum) was taken and is essential for understanding amplitude measurements (here we have the relation to mb). In principle the scaling time is contained in the amplitude time window as reference time but the more appropriate QuakeML attribute for this is the scaling time.

### DIFF
--- a/libs/seiscomp/processing/amplitudes/iaspei.cpp
+++ b/libs/seiscomp/processing/amplitudes/iaspei.cpp
@@ -1,0 +1,150 @@
+#include <seiscomp/processing/amplitudes/iaspei.h>
+
+#include <vector>
+#include <complex>
+#include <cmath>
+
+#include <seiscomp/math/filter/seismometers.h>
+
+namespace Seiscomp {
+
+namespace Processing {
+
+namespace IASPEI {
+
+#include <seiscomp/math/minmax.ipp>
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+bool measureAmplitudePeriod(
+	std::size_t n,
+	const double *f,
+	double offset,
+	std::size_t istart,
+	std::size_t iend,
+	AmplitudePeriodMeasurement &measurement)
+{
+	// NOTE:
+	// This routine scans the input data array twice, once for the zero
+	// crossings and once for the peaks (incl. troughs!) in between.
+	// This makes the routine easier to understand but at some
+	// computational cost that in tests have shown to be irrelevant in
+	// a normal processing setup.
+
+	if ( istart >= iend || iend >= data.size() ) {
+		return false;
+	}
+
+	// Determine zero crossings.
+	std::vector<std::size_t> zero_crossings;
+	double previous { f[istart] - offset };
+	for ( std::size_t i = istart + 1; i < iend; i++ ) {
+		double current { f[i] - offset };
+		if ( current*previous <= 0 && previous != 0 ) {
+			// => zero crossing
+			// i refers to the first sample after zero crossing
+			zero_crossings.push_back(i);
+		}
+
+		previous = current;
+	}
+
+	if ( zero_crossings.size() < 3 ) {
+		return false;
+	}
+
+	// Determine peaks between zero crossings.
+	std::vector<std::size_t> peaks;
+	std::size_t nz { zero_crossings.size() };
+	for ( std::size_t k = 1; k < nz; k++ ) {
+		// The casts will never hurt in practice but should be
+		// removed after a revision of find_absmax().
+		int i1 = static_cast<int>(zero_crossings[k-1]);
+		int i2 = static_cast<int>(zero_crossings[k]);
+		std::size_t peak = find_absmax(static_cast<int>(n), f, i1, i2, offset);
+		peaks.push_back(peak);
+	}
+
+	// Keep track of the largest p2p amplitude...
+	double amax { 0 };
+	// ... and the respective index
+	std::size_t kmax { 0 };
+
+	std::size_t np = peaks.size();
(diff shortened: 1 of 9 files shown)


libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+namespace Seiscomp {
+
+namespace Processing {
+
+namespace IASPEI {
+
+
+static std::size_t
+find_absmax(std::size_t n, const double *f, double offset, std::size_t i1, std::size_t i2)
```
gempa-jabe, line 15: This method is already available in `math/minmax.ipp`. Is it different to this one?

libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+	// This makes the routine easier to understand but at some
+	// computational cost. Tests have shown that on a normal laptop
+	// (as of 2023) this routine is capable of making more than 100k
+	// measurements per second, so it will hardly be a bottleneck in
+	// a normal processing setup.
+
+	// Determine zero crossings.
+	std::vector<std::size_t> zero_crossings;
+	double previous { f[istart]-offset };
+	for (std::size_t i = istart+1; i < iend; i++) {
```
gempa-jabe, line 52: Spaces between parenthesis, see https://www.seiscomp.de/doc/base/coding-conventions.html.

libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+namespace Processing {
+
+namespace IASPEI {
+
+
+static std::size_t
+find_absmax(std::size_t n, const double *f, double offset, std::size_t i1, std::size_t i2)
+{
+	double fmax = std::abs(f[i1]-offset);
```
gempa-jabe, line 17: I recommend to use spaces around operator to improve readability.

libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+			imax = i;
+			fmax = fabs;
+		}
+	}
+
+	return imax;
+}
+
+
+bool measure_amplitude_period(
```
gempa-jabe, line 32: Function names with camelCase `measureAmplitudePeriod(...)`.

libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+double wwssnsp_amplitude_response(double frequency_hz)
+{
+	// Source:
+	// - Jim Dewey (NEIC)
+	// - IASPEI Magnitude WG.
+	// - Havskov and Ottemoeller (2010)
+	//   DOI: 10.1007/978-90-481-8697-6_6
+	//   Table 6.7
+
+	static std::vector<std::complex<double>> poles {
```
gempa-jabe, line 137: Wouldn't it be better to expose this vector in https://github.com/SeisComP/common/blob/master/libs/seiscomp/math/filter/seismometers.cpp#L392 and reuse it here?

gempa-jabe: Actually doing

```cpp
static WWSSN_SP_Filter wwssn_sp(Dispacement);

wwssn_sp.poles;
wwssn_sp.zeros;
wwssn_sp.norm;
```

would do the job in a less redundant way.

libs/seiscomp/processing/amplitudes/iaspei.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+
+bool measure_amplitude_period(
+	const std::vector<double> &data,
+	double offset,
+	std::size_t istart,
+	std::size_t iend,
+	AmplitudePeriodMeasurement &measurement)
+{
+	if (istart >= iend || iend >= data.size())
```
gempa-jabe, line 118: Using curly brackets also for one line blocks is recommended.

libs/seiscomp/processing/amplitudes/m_B.cpp
```diff
@@ -62,6 +63,24 @@ AmplitudeProcessor_mB::AmplitudeProcessor_mB(const Core::Time& trigger)
 
 
 
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+void AmplitudeProcessor_mB::finalizeAmplitude(DataModel::Amplitude *amplitude) const {
+	if ( amplitude == NULL )
```
gempa-jabe, line 69: Use `!amplitude`.

gempa-jabe (changes requested):
Copyright headers are missing in the new files.

gempa-jabe (approved):
Just note that a port to API 15 is not possible due to the structure layout changes in `AmplitudeProcessor::Config`.

gempa-jabe:
@jsaul, give me a go and I will merge it.

PR #91, opened 2023-04-11
[stdloc] add confidence ellipsoid computation

luca-s:
I have finally managed to add the computation of the confidence ellipsoid.


plugins/locator/stdloc/stdloc.h
```diff
@@ -77,6 +77,28 @@ class StdLoc : public Seiscomp::Seismology::LocatorInterface {
 	//  Private members
 	// ----------------------------------------------------------------------
 	private:
+
+		// Covariance Matrix (X axis as W-E and Y axis as S-N)
+		// Since it is a diagonal matrix the duplicated values
+		// are commented
+		struct CovMtrx {
+			bool valid;
+		  double   sxx,  sxy,  sxz,  sxt;
```
gempa-jabe, line 86: Something wrong with tab and spaces. 3 tabs here ...

luca-s:
Fixed. Do you actually use an automatic formatting tool? So that I can use the same and avoid this little mistakes.

gempa-jabe:
I have been working on a clang-format config file which I pushed to the seiscomp repository in `.clang-format`. It is far from being complete and well tested but at least it is a start. Unfortunately it swallows the empty lines between the function spacers.

PR #92, opened 2023-04-11
[io] add support for wildcards in File RecordStream

luca-s:
File::addStream now supports * ? () |

This is related to https://github.com/SeisComP/main/pull/59 - without wildcards support that PR is of limited use.

### DIFF
--- a/libs/seiscomp/io/recordstream/file.cpp
+++ b/libs/seiscomp/io/recordstream/file.cpp
@@ -24,6 +24,7 @@
 #include <seiscomp/logging/log.h>
 #include <seiscomp/system/environment.h>
 
+
 using namespace std;
 using namespace Seiscomp;
 using namespace Seiscomp::RecordStream;
@@ -145,10 +146,10 @@ bool File::addStream(const string &net, const string &sta,
                      const string &loc, const string &cha) {
 	string id = net + "." + sta + "." + loc + "." + cha;
 	if ( id.find_first_of("*?") == std::string::npos ) {
-		_filter[id] = TimeWindowFilter();
+		_filter.emplace(id, TimeWindowFilter());
 	}
 	else { // wildcards characters are present
-		_reFilter.push_back(make_pair(id, TimeWindowFilter()));
+		_reFilter.emplace_back(id, TimeWindowFilter());
 	}
 	return true;
 }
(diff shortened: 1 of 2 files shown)


gempa-jabe:
This will break the ABI and could only be used with SC >= 6. I have had decided against stream ID filtering in the file RS for performance reasons. Furthermore full regexp support is not in SDS archive and I would prefer to keep both filter options in sync. That is why `wildcmp` should be used rather than regexp. The current implementation only filters records if the streamID is wildcarded otherwise no filter is in place. If that is correct, does that make sense? Adding `XX.ABCD..BHZ` would not have an effect?

If all that is really required to make `scart` work we need to carefully design the file rs to not break existing code.

luca-s:
> This will break the ABI and could only be used with SC >= 6. 

I am aware of it and it is not an issue.

>  I have had decided against stream ID filtering in the file RS for performance reasons. 

I thought about it while modifying the code and this PR doesn't slow down the previous behaviour of not having wildcards. So the performance hit, if there is a hit, can be seen only if the wildcards are in use. But do you really think there could be a problem with performance? I don't see a real world use case where this can be a problem, but I can see a  couple of scenarios where this feature might be useful. However I don't want to insist. Just some thoughts.

>  Furthermore full regexp support is not in SDS archive and I would prefer to keep both filter options in sync. That is why wildcmp should be used rather than regexp. 

Good point I can change it. I don't believe loosing the ability to make use of ( ) and | is a big deal. 

> The current implementation only filters records if the streamID is wildcarded otherwise no filter is in place. 

Are you talking about wildcmp or File rs now?


> If all that is really required to make scart work we need to carefully design the file rs to not break existing code.

This change is not required, `scart` can work without it, but the user would only be able to filter with fully qualified stream codes, that is without wildcards. However, since `scart` was never able to use stream/time filtering with File as input, then I don't see any problem with that.

gempa-jabe:
> Good point I can change it. I don't believe loosing the ability to make use of ( ) and | is a big deal.

And if we need it anytime in future it could be added to `wildcmp`.

> Are you talking about wildcmp or File rs now?

The file rs. I see that I did a mistake interpreting the code. I think using `wildcmp` will be fine and if carefully implemented will not hit the performance as you said.

gempa-dirk:
> This change is not required, `scart` can work without it, but the user would only be able to filter with fully qualified stream codes, that is without wildcards. However, since `scart` was never able to use stream/time filtering with File as input, then I don't see any problem with that.

I found the stream and time filtering useful when generating SDS from files and I would be happy if this new feature could be kept. Otherwise, the intermediate step via scmssort is required but of course doable.

libs/seiscomp/io/recordstream/file.cpp
```diff
@@ -211,13 +221,44 @@ bool File::setRecordType(const char *type) {
...
+	const auto & it = _filter.find(rec->streamID());
+	if ( it != _filter.end() ) {
+		return &it->second;
+	}
+
+	// then search the wildcarded filters
+	for ( const auto& pair : _reFilter ) {
+		const string& wild = pair.first;
+		const TimeWindowFilter& twf = pair.second;
+		if ( Core::wildcmp(wild, rec->streamID()) ) {
```
gempa-jabe, line 237: `rec->streamID()` is rather expensive as it composes the string on demand. Please save this value at the top of the function. It should be called only once.

gempa-jabe:
The code is looking good to me and I would merge it. Any blockers?

gempa-jabe:
I have tested it and it worked. Thanks!

PR #95, opened 2023-06-27
[logging] avoid declaration of 'expect' macro

luca-s:
Calling a macro with such a common word (expect) easily results in code breakage. The macro is not even useful (it might have been useful in the past).

I would suggest to completely remove the macro.

### DIFF
--- a/libs/seiscomp/logging/common.h
+++ b/libs/seiscomp/logging/common.h
@@ -40,10 +40,8 @@
 */
 #ifdef __GNUC__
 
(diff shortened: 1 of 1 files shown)


luca-s:
In my use case the macro breaks the compilation of code that uses boost json. 

image

gempa-jabe:
OK, this can be removed. We have already rapidjson included in the common package as 3rd-party lib. Maybe it could be an alternative to another json parser. But that is completely independent of this PR.

luca-s:
thanks

PR #97, opened 2023-07-13
[quakeml] fix invalid ResourceUri

andres-h:
When using obspy.client to retrieve data there is a user warning for each arrival. The arrivals in quakeml-format do not have a valid publicID.

UserWarning: 'smi:org.gfz-potsdam.de/geofon/smi:local/LED/pick/6b98ba97-c5b1-4488-ae7c-29752b40e0fb_smi_local/LED/c4c35d6d-cd06-4596-bcd8-0956f155c935' is not a valid QuakeML URI. It will be in the final file but note that the file will not be a valid QuakeML file.


andres-h:
Yes, sorry, I admit that I did not test, because the change seemed trivial and the problem was reported by another user.

PR #98, opened 2023-07-28
[io] Fix concurrent recordstream termination when data still available

luca-s:
I found out that [Concurrent::next()](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.cpp#L271) can return `nullptr` even if there is still data in the [Concurrent::_queue](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.h#L113) and thus the requested data time window is not fully returned.

When a recordstream thread terminates, [it closes the queue](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.cpp#L261) and that forces the next call to [_queue.pop()](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.cpp#L300) to terminate via [QueueClosedException.](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.cpp#L303) . That happens even when there is still data in the queue.

So I changed the behaviour of [ThreadedQueue](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/client/queue.h) so that it returns the remaining items via `Concurrent::pop()` before raising a `QueueClosedException`. That is, no more items are accepted in a closed queue via `Concurrent::push`, but it is still possible to call `Concurrent::.pop()` until the queue is empty.

 I don't know if some code relies on the current behaviour of `ThreadedQueue::pop`. If that is the case, then I need to find an alternative fix for concurrent recordstream

### DIFF
--- a/libs/seiscomp/io/recordstream/concurrent.cpp
+++ b/libs/seiscomp/io/recordstream/concurrent.cpp
@@ -257,8 +257,8 @@ void Concurrent::acquiThread(RecordStream *rs) {
 
 	_mtx.lock();
 	if ( --_nthreads == 0) {
-		SEISCOMP_DEBUG("Closing record queue");
-		_queue.close();
+		SEISCOMP_DEBUG("Last acquisition thread terminated");
+		_queue.push(nullptr);
 	}
 	_mtx.unlock();
 }
@@ -297,9 +297,22 @@ Record *Concurrent::next() {
 	}
 
 	try {
-		return _queue.pop();
+		Record * rec =  _queue.pop();
+		if ( rec) {
+			return rec;
+		}
+
+		if ( _queue.size() > 0 ) {
+			SEISCOMP_ERROR("Finished acquisition, but data queue not empty "
+			               "(%zu items still)", _queue.size());
+		}
+
+		SEISCOMP_DEBUG("Closing record queue");
+		_queue.close();
 	}
 	catch ( Client::QueueClosedException & ) {
+		// Since the queue was closed,it doesn't matter if there are still items
+		// in the queue, we drop them
 		SEISCOMP_DEBUG("Queue closed, break streaming");
 	}
 


gempa-jabe:
Hi Luca, thank you for bringing this issue on the table. I must admit that I do not like the fix. If a queue is closed then it means closed, regardless of how many items are still queued.

To synchronize the flow in the concurrent rs, we could just send a nullptr to signal the end of acquisition of all threads. This could in turn also cause a call to `ThreadedQueue::close` (if necessary) in `Concurrent::next()`. What do you think?

luca-s:
I suspected you didn't like the change on the `ThreadedQueue` api behaviour, because it changes the semantics and I am not a big fan either on this type of changes.

However, I wondered when would the current behaviour be useful? Depleting the queue before closing it on the pop side has advantages but I do not see disadvantages, that's why I proposed the change anyway.

> To synchronize the flow in the concurrent rs, we could just send a nullptr to signal the end of acquisition of all threads.  This could in turn also cause a call to ThreadedQueue::close (if necessary) in Concurrent::next(). What do you think?

I feel we are doing a step back here, I do not like it much because we are adding meaning to the return value of pop instead of relying on the `ThreadedQueue`.

Other ways of fixing the issue could be to extend the `ThreadedQueue` API:
- option 1: support a new method of closing only the push side (it is quite common for a threaded queue to close only the write side first and let the read side continues until the end of the data)
- option 2: add a method that allows to fetch the remaining items after the queue has been closed (and raises an exception if the queue is not closed)

If you prefer to use your approach (the nullpt) it's ok with me though.

gempa-jabe:
> Depleting the queue before closing it on the pop side has advantages but I do not see disadvantages

This class is being used in other codes which might have much more items than just 10, maybe several thousands. If you close something then you do not want to work anymore with it, e.g. you do not want to process another 1000 queued items. It actually delays e.g. closing a window which is using this queue.

> option 2: add a method that allows to fetch the remaining items after the queue has been closed (and raises an exception if the queue is not closed)

This would be the option I *would* prefer but it adds more checks and conditional statements to the `next` method. Without wanting to push my idea further, I just want to add that the return value of `next` has exactly the same meaning. `nullptr` means the end of data. No additional control flow required.

luca-s:
Ok you convinced me, both solutions have advantages. If you don't mind I let you decide which one you prefer and I will implement and test it since I have a reproducible bug ready on my computer. 

**Option 1**:  return `nullptr` instead of [closing the queue](https://github.com/SeisComP/common/blob/4422e4418da5266cacb3e471fd6391bae313d898/libs/seiscomp/io/recordstream/concurrent.cpp#L261).

Then `Concurrent::next()` becomes:
```
	try {
		Record * rec =  _queue.pop();
                if ( !rec) {
                  _queue.close();
                }
                return rec;
	}
	catch ( Client::QueueClosedException & ) {
		SEISCOMP_ERROR("Internal logic error: queue closed, but this should never happen");
	}
        return nullptr;
```

**Option 2**: we add the method `ThreadedQueue::deplete()`

Then `Concurrent::next()` becomes:
```
	try {
		return _queue.pop();
	}
	catch ( Client::QueueClosedException & ) {
	       SEISCOMP_DEBUG("All threads terminated, depleting queue");
	}
	return _queue.deplete();
```

luca-s:
I like your implementation. So I will try this and If we discover any issue that we haven't thought about yet, we can fall back to the option 1.

luca-s:
It turned out I prefer your initial idea (option 1). This is because we can distinguish the case where the queue was intentionally closed (e.g. form a GUI) - in that case we can drop any remaining item in it - and the case where the acquisition threads all terminated - in this case we want to return until the last item -. 

I tested the code and it works fine. Please let know know if there are any additional changes you would like to add.

libs/seiscomp/io/recordstream/concurrent.cpp
```diff
@@ -297,9 +297,22 @@ Record *Concurrent::next() {
 	}
 
 	try {
-		return _queue.pop();
+		Record * rec =  _queue.pop();
+		if ( rec) {
+			return rec;
+		}
+
+		if ( _queue.size() > 0 ) {
```
gempa-jabe, line 305: That must never happen. We push `nullptr` after the last thread terminated. No way that there are more items in the queue.

gempa-jabe: Anyway, we can keep it for safety and monitor the logs.

gempa-jabe:
If you are fine with the implementation, I will merge it.

luca-s:
yes please, go ahead and merge it.

libs/seiscomp/io/recordstream/concurrent.cpp
luca-s: I added that check to avoid regressions in the future. However I was tempted to drop it too.

PR #100, opened 2023-08-11
[locator] fix possible infinite loop in locator uncertainty computation

luca-s:
This bug affects `LOCSAT` and `StdLoc`.

I noticed this infinite loop because `StdLoc` enables the confidence ellipsoid by default (LOCSAT has it disabled). Since I am working on  playbacks with tens of thousands of events the bug became apparent.

I don't know why they implemented the original code in that way, but they must have had a good reason. So I am not super happy to change the implementation, but the possible infinite loop is way too problematic. I'll let you decide what to do.

Here is a naive test for the changed code, in case you want to play around with it:

```
#include <iostream>
#include <cmath>

using namespace std;

/*
 * Code copied from common/libs/seiscomp/seismology/locator/eigv.cpp
 */
double r8_max ( double x, double y )
{
  double value;

  if ( y < x )
  {
    value = x;
  }
  else
  {
    value = y;
  }
  return value;
}

double r8_min ( double x, double y )
{
  double value;

  if ( y < x )
  {
    value = y;
  }
  else
  {
    value = x;
  }
  return value;
}

double r8_abs ( double x )
{
  double value;

  if ( 0.0 <= x )
  {
    value = + x;
  }
  else
  {
    value = - x;
  }
  return value;
}

double pythag ( double a, double b )
{
  double p;
  double r;
  double s;
  double t;
  double u;

  p = r8_max ( r8_abs ( a ), r8_abs ( b ) );

  if ( p != 0.0 )
  {
    r = r8_min ( r8_abs ( a ), r8_abs ( b ) ) / p;
    r = r * r;

    while ( 1 )
    {
      t = 4.0 + r;

      if ( t == 4.0 )
      {
        break;
      }

      s = r / t;
      u = 1.0 + 2.0 * s;
      p = u * p;
      r = ( s / u ) * ( s / u ) * r;
    }
  }
  return p;
}

/*
 * Simple test to compare the 2 implementations
 */
void compare(double percentDiff, double rangeA, double rangeB, double stepA, double stepB)
{
  for (double a = -rangeA; a <= rangeB; a +=stepA)
    for (double b = -rangeB; b <= rangeB; b +=stepB)
    {
      double ref = pythag(a,b);
      double alt = sqrt(a*a+b*b);
      double diff = ref-alt;
      if ( fabs(diff/ref) > percentDiff )
      {
        cout << diff << " " << ref << " " << alt << endl;
      }
    }
}

int main()
{
  const double percentDiff = 0.000000000000001;

  compare(percentDiff, 1e20, 1e20, 1e17, 1e17);
  compare(percentDiff, 1e10, 1e10, 1e7, 1e7);
  compare(percentDiff, 1e3, 1e3, 1, 1);

  compare(percentDiff, 1, 1, 1e-3, 1e-3);
  compare(percentDiff, 1e-7, 1e-7, 1e-10, 1e-10);
  compare(percentDiff, 1e-17, 1e-17, 1e-20, 1e-20);

  compare(percentDiff, 1e20, 1e-17, 1e17, 1e-20);
  compare(percentDiff, 1e10, 1e-7, 1e7, 1e-10);
  compare(percentDiff, 1e3, 1, 1, 1e-3);
  return 0;
}


```

### DIFF
--- a/libs/seiscomp/seismology/locator/eigv.cpp
+++ b/libs/seiscomp/seismology/locator/eigv.cpp
@@ -18,6 +18,7 @@
  ***************************************************************************/
 #include <iostream>
 #include <math.h>
+#include <limits>
 using namespace std;
 #include "eigv.h"
 
@@ -328,6 +329,10 @@ double pythag ( double a, double b )
   double s;
   double t;
   double u;
+
+  if ( !std::isfinite(a) || !std::isfinite(b) ) {
+     return std::numeric_limits<double>::quiet_NaN();
+  }
 
   p = r8_max ( r8_abs ( a ), r8_abs ( b ) );
 


gempa-jabe:
Why a special version of pythag is implemented, is outlined in the comment. Why it is implemented as it is, is not and I don't know the background of the implementation. An infinite loop is not acceptable and we need to fix it. Your approach is reasonable until we know better. I would even throw an exception if an overflow occurs, e.g. after checking of nan of the result, but that has to be checked with a test. In that case the eigenvalues cannot be computed. I would furthermore only export required functions and remove all the internal `r8_*` methods from the export. Actually only `rs` is required and that can go into the math library.

Actually we have a tensor class (`Tensor2S`: 3d symmetric matrix) which can compute the eigenvalues and which is used by the MT codes. This should work as a replacement as we currently only compute eigenvalues from 2 and 3 matrices.

luca-s:
> 2. Fix the bug in pythag

This seems to be the nicest option. Let me see if I can reproduce the infinite loop and give you more details.

luca-s:
image

Well, it looks like the function was called with `-nan ` as initial value. So we could simply return `nan` if one of the parameters is not defined and that would solve the issue (not all maybe).

gempa-jabe:
This makes sense so let's merge it.

PR #104, opened 2023-09-11
[MLc] fix computation for distanceMode = 'hypocentral'

luca-s:
The current implementation considers the origin depth as the vertical distance to the receiver, which is an assumption that breaks at local scale. The fix computes the vertical distance as the difference between sensor elevation and origin depth.

### DIFF
--- a/libs/seiscomp/processing/magnitudes/MLc.cpp
+++ b/libs/seiscomp/processing/magnitudes/MLc.cpp
@@ -24,7 +24,7 @@
 #include <seiscomp/processing/magnitudes/MLc.h>
 #include <seiscomp/config/config.h>
 #include <seiscomp/logging/log.h>
-
+#include <seiscomp/client/inventory.h>
 
 using namespace std;
 
@@ -217,7 +217,7 @@ bool MagnitudeProcessor_MLc::initLocale(Locale *locale,
 MagnitudeProcessor::Status MagnitudeProcessor_MLc::computeMagnitude(
 	double amplitude, const std::string &unit,
 	double, double, double delta, double depth,
-	const DataModel::Origin *, const DataModel::SensorLocation *,
+	const DataModel::Origin *, const DataModel::SensorLocation *receiver,
 	const DataModel::Amplitude *,
 	const Locale *locale,
 	double &value) {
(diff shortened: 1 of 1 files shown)


libs/seiscomp/processing/magnitudes/MLc.cpp
```diff
@@ -264,12 +264,20 @@ MagnitudeProcessor::Status MagnitudeProcessor_MLc::computeMagnitude(
...
 
 	SEISCOMP_DEBUG("  + distance mode: %s", distanceMode.c_str());
+
+	double hDistanceKm = Math::Geo::deg2km(delta);
+	double vDistanceKm = 0;
+	if ( !receiver && distanceMode == "hypocentral" ) {
+		return MetaDataRequired;
+	}
+	if ( receiver ) {
+		vDistanceKm = abs(receiver->elevation() / 1000. + depth);
```
gempa-dirk, line 274: maybe abs() is not required here and it even would hide the information if the event is below or  above the sensor  which is also reported in debug output.

gempa-dirk:
The changes are reasonable.

libs/seiscomp/processing/magnitudes/MLc.cpp
luca-s: Good point.

libs/seiscomp/processing/magnitudes/MLc.cpp
```diff
@@ -264,12 +264,20 @@ MagnitudeProcessor::Status MagnitudeProcessor_MLc::computeMagnitude(
...
+		vDistanceKm = receiver->elevation() / 1000. + depth;
+	}
 	double distanceKm;
 	if ( distanceMode == "hypocentral" ) {
-		distanceKm = sqrt(pow(Math::Geo::deg2km(delta),2) + pow(depth,2));
-	}
-	else {
-		distanceKm = Math::Geo::deg2km(delta);
+		distanceKm = sqrt(pow(hDistanceKm, 2) + pow(vDistanceKm,2));
+	} else {
```
gempa-jabe, line 279: FYI: we want a newline before `else`. I am going to fix that after merging.

Issue #80, opened 2023-02-08
Should PickerView and AmplitudeView consider the hypocenter distance?


luca-s:
I  am working on two projects where we have several boreholes with multiple sensors per borehole. In this scenario, when I use the PickerView or the AmplitudeView the "sorting by distance" feature doesn't work well because the GUIs consider only the epicenter distance instead of the hypocenter distance and the events are so close to the sensors that the difference in their elevation matters.

I was considering providing a PR  to fix this, but after reviewing the code of PickerView and AmplitudeView  I have the feeling that they were never meant to be used for such local processing  where the hypocenter vs epicenter distance is relevant. 

Do you think it would make sense to fix this use case?

gempa-jabe:
If so how do you think that both components were never meant to support your request? Any hints in the code that are an show stopper?

luca-s:
Hi @gempa-jabe.  I still believe this could be an interesting enhancement for SeisComP, as long as you are interested in the usage of SeisComP at low scale. 

However if we want to tackle the problem in a consistent way there are a couple of places in the code where the epicenter vs hypocenter distance plays a role that we need to fix:
- PickerView: this is the main reason that made me open this issue. At the moment I run a locally modified version of the code to deal with the sorting issue on the PickerView. The patch is simple but the distance could be displayed in degree (instead of km), so it is pretty obvious that it represents epicenter distance.
- AmplitudeView: how the distance is handled here is similar to the `PickerView`. However there is an additional detail: the computed distance is passed to the `magnitudeProcessor` when calling `computeMagnitude` and changing the code would result in different magnitude values (one might argue the change is towards more correct values).
- scolv `Distance`,`RravelTime` and `MoveOut` plots: the epicenter distance is considered and not the hypocenter distance. The distance information is computed by the locator and all locators (except for stdloc)  compute the epicenter distance. 

Overall, I believe SeisComP would perfectly work at low scale since I am using it for that purpose, but the code might need some change here and there to remove the current assumptions that we are working at regional scale and station elevations do not matter.

luca-s:
(I made few edits to my previous post)

gempa-jabe:
As this feature is gaining more momentum currently, I would suggest the following approach to this:
* Configure hypocentral distance with `scheme.distanceHypocentral = true`.
* Compute distances according to this setting: PickerView, AmplitudeView, OriginLocatorView
[...]

PR #105, opened 2023-09-11
[pickerview] sort stations by hypocentral distance instead of epicentral one

luca-s:
This complete part of #80

Please note that only the sorting has been fixed. The maximum station distance selector still refers to epicenter distance. There are too many hassles to deal with if we want to change that behaviour too and I don't believe is required at all.

### DIFF
--- a/libs/seiscomp/gui/datamodel/pickerview.cpp
+++ b/libs/seiscomp/gui/datamodel/pickerview.cpp
@@ -4971,6 +4971,9 @@ bool PickerView::addTheoreticalArrivals(RecordViewItem* item,
 		double delta, az1, az2;
 		double elat = SC_D.origin->latitude();
 		double elon = SC_D.origin->longitude();
+		double edep = SC_D.config.defaultDepth;
+		try { edep = SC_D.origin->depth(); } catch ( ... ) {}
+
 		double slat, slon;
 		double salt = loc->elevation();
 
@@ -4984,8 +4987,12 @@ bool PickerView::addTheoreticalArrivals(RecordViewItem* item,
 		}
 
 		Math::Geo::delazi(elat, elon, slat, slon, &delta, &az1, &az2);
+		double hDist = Math::Geo::deg2km(delta);           // [km]
+		double vDist = edep + salt/1000.;                  // [km]
+		double distance = sqrt(hDist*hDist + vDist*vDist); // [km]
+		distance = Math::Geo::km2deg(distance);            // [degree]
 
-		item->setValue(ITEM_DISTANCE_INDEX, delta);
+		item->setValue(ITEM_DISTANCE_INDEX, distance);
 		item->setValue(ITEM_AZIMUTH_INDEX, az1);
 
 		/*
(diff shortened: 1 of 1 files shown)


libs/seiscomp/gui/datamodel/pickerview.cpp
```diff
@@ -4971,6 +4971,7 @@ bool PickerView::addTheoreticalArrivals(RecordViewItem* item,
 		double delta, az1, az2;
 		double elat = SC_D.origin->latitude();
 		double elon = SC_D.origin->longitude();
+		double edep = SC_D.origin->depth();
```
gempa-jabe, line 4974: This is optional and can throw an exception.

gempa-jabe:
I am fine with the implementation if the depth issue is resolved. I think there is a default depth configured if the origin depth is not set.

libs/seiscomp/gui/datamodel/pickerview.cpp
luca-s: Ah! I was not aware of that. I will fix this

luca-s:
PR updated

luca-s:
As I side note, I  have thought of applying the same logic of this PR to the `amplitudeview`, that is to populate the `ITEM_DISTANCE_INDEX` value of `RecordViewItem` with the hypocentral distance instead of the epicentral one. That would result in the correct sorting of the stations by hypocentral distance. 

However I cannot do that because in the `amplitudeview` the `ITEM_DISTANCE_INDEX` value of `RecordViewItem` is also used to store the **epicentral** distance to be passed to the magnitude processor (see [here](https://github.com/SeisComP/common/blob/ac5b4d432ad2018354afe3e925b83088effdc077/libs/seiscomp/gui/datamodel/amplitudeview.cpp#L2721C24-L2721C24) and [here](https://github.com/SeisComP/common/blob/ac5b4d432ad2018354afe3e925b83088effdc077/libs/seiscomp/gui/datamodel/amplitudeview.cpp#L3322)  for example). 

I am mentioning this because you might have an idea of a simple workaround that doesn't require a major code change.

gempa-dirk:
@luca-s The solution is a sound approximation at short distance where straight rays may be assumed and it may be reasonable at local to regional distances as weöö. At larger distances a hypocentral distance cannot be estimated by the cartesian distance but should it be the length of a ray? On the other hand the current implementation assumes epicentral distance and vertical distance hence, it may not harm. At larger epicentral distances will be longer than cartesian hypocentral distances. This will totally change the sorting of the traces resulting in very uncommon travel-time curves.

Therefore, a clear definition should be given somewhere, e.g. in the glossary, what a hypocentral distance actually is in a particular context. Note, some applications do may consider cartesian distances.

luca-s:
I understand your comment, but at larger distances, when my approximation breaks, the epicentral distance is so large that it doesn't matter the vertical one, so the sorting would still be correct.

That said , an improvement would be to consider the hypocentral distance only when the epicentral distance is smaller than a certain value. Even better, only when the vertical distance is greater than X% of the epicentral one. What about it?

luca-s:
Let's drop this PR then. This should be a simple fix transparent to the user. If it gets too complicated I don't like it.

I can personally live with my locally modified version of the code for the few projects I maintain, so I don't necessarily need this PR.

Just be aware that if more users start using seiscomp at local scale then #80 becomes relevant.

Issue #109, opened 2023-09-14
Proposal: updating the TravelTimeTableInterface


luca-s:
Dear developers,

I would like to suggest a couple of changes to the `TravelTimeTableInterface` and hear your opinions on that. 

1 - add a `TravelTimeTableInterface::computeTravelTime` method that returns only the travel time without additional information (no `TravelTime` is returned). This is because the travel time is the most requested information from the `TravelTimeTableInterface` throughout seiscomp and this change would allow to improve the overall performance, since returning only the travel time can potentially be less expensive than returning all the fields of the `TravelTime` class.

2 - add a field `azimuth` to the `TravelTime` class. This is to allow 3D velocity models in the future. In a 3D velocity model the azimuth of the take off angle is calculated by the travel time table, because the ray horizontal path can be reflected/refracted/bent the same way it happens with the vertical ray path on a 1D model.  This change would make the existing 1D models return the azimuth computed by e.g. delazi. 

What are your thoughts on this?

thanks,
Luca

gempa-jabe:
I am fine with both suggestions. The azimuth can also be set to undefined in current models.

luca-s:
You mean like setting the azimuth to nan? That seems interesting, I didn't think about it. It might have some advantages, like properly communicating that value is not handled by the ttt. On the other hand, returning a value, e.g. from delazi, centralize at api level the precision of the azimuth computation. I believe both solutions have good points. 

Either way, if you agree, I will provide a PR for this.

gempa-jabe:
`nan` would be an option, indeed. I had `OPT(double)` in mind. If the `delazi` computation does not add too much computational overhead to the default implementation, I am also fine with it.

gempa-jabe:
Please create your PR and we can continue to discuss the code.

PR #111, opened 2023-09-21
[ttt] update api w.r.t. issue #109 

luca-s:
This is the general idea I had in mind for the issue #109. At the moment it doesn't seem a huge performance improvement, but I am thinking at future TTT types that i would like to add.

### DIFF
--- a/libs/3rd-party/locsat/lib/libloc/locate_event.c
+++ b/libs/3rd-party/locsat/lib/libloc/locate_event.c
@@ -1058,7 +1058,7 @@ int find_phase(const char *phase) {
 }
 
 
-double compute_ttime(double distance, double depth, char *phase, int extrapolate,
+double compute_ttime(double distance, double depth, const char *phase, int extrapolate,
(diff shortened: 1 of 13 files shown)


luca-s:
@gempa-jabe this is not urgent, but I just wanted to be sure you didn't miss this PR. Thanks.

gempa-jabe:
I have checked it and it is looking good to me. I am just wondering if `computeTime` is a bit shorter and descriptive enough as we are already in the context of computing travel times.

luca-s:
Thank you @gempa-jabe. `computeTime` alternative sounds good to me. I will update the PR and rebase on `origin/master` (I'll drop your commit since I need to apply EXTRAPOLATE to Locsat::computeTravelTime too).

gempa-jabe:
Can you squash the lastest three commits or squash the rename commit into the previous two as all the changes are part of a single PR and this should only contain the final code and no evolution.

luca-s:
Sure I can, done ;)

gempa-jabe:
Great, so let's merge it.

PR #113, opened 2023-10-24
Update ferdata.h

comoglu:
Change the Country name from fep file as the country name officially changed.


gempa-jabe:
Is it really "Turkiye" or "Türkiye"? From what I have checked quickly, it is spelled with ü.

comoglu:
Yes It is indeed, "Türkiye" But I was not sure if it breaks anything with c code. If "Türkiye" acceptable. I can update the pull request

gempa-jabe:
Thanks. Let's see if it will break anything. The file must be encoded with utf-8 then everything should be fine.

comoglu:
It worked fine and broke nothing. 
image

PR #118, opened 2023-11-15
[io] concurrent record stream clean up

luca-s:
I would like to open this PR for a discussion on some minor changes to the concurrent RS.

A - [protecting `--_nthreads` with a mutex ](https://github.com/SeisComP/common/blob/be54f7201ea22381a717658476b5e43ca3e4f065/libs/seiscomp/io/recordstream/concurrent.cpp#L302) is not required, because `_nthreads` is only used within `Concurrent::next`. This is the change that I care most, because it impacts the performance

B - `Concurrent::close()` should be performed [independently of the value of `_started` ](https://github.com/SeisComP/common/blob/be54f7201ea22381a717658476b5e43ca3e4f065/libs/seiscomp/io/recordstream/concurrent.cpp#L207), so that  it is possible to release resources on a RS that was never started (no call to  `Concurrent::next()` ).

C - Setting `_started=false` [at the end](https://github.com/SeisComP/common/blob/be54f7201ea22381a717658476b5e43ca3e4f065/libs/seiscomp/io/recordstream/concurrent.cpp#L229) of `Concurrent::close()` doesn't feel right because a record stream cannot be restarted once closed.  Leaving `_started` as `true` allows a cleaner implementation of `Concurrent::next()` since it doesn't have to handle the case of a closed/reopened RS. It is also a good flag to indicates that a RS has terminated its job

Please let me know if you like the proposed changes and I will perform a thorough testing before merging.

### DIFF
--- a/libs/seiscomp/io/recordstream/concurrent.cpp
+++ b/libs/seiscomp/io/recordstream/concurrent.cpp
@@ -203,12 +203,7 @@ bool Concurrent::setTimeout(int seconds) {
 // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 void Concurrent::close() {
 	lock_guard<mutex> lock(_mtx);
-
-	if ( !_started ) {
-		SEISCOMP_DEBUG("Closing without being started");
-		return;
-	}
-
+	
 	_queue.close();
 
 	for ( size_t i = 0; i < _rsarray.size(); ++i ) {
@@ -220,13 +215,9 @@ void Concurrent::close() {
 			thread.join();
 		}
 	}
-
-	SEISCOMP_DEBUG("All acquisition threads finished");
-
+	
 	_threads.clear();
 	_rsarray.clear();
-
-	_started = false;
 }
 // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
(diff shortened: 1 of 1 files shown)


gempa-jabe:
Thanks for digging again into it. @andres-h has had some issues we had to fix recently. It was related to closing from within another thread in scolv iirc. Any additional changes or fixes must be tested indeed carefully to not break it again. The issue exhibited with the balanced rs.

luca-s:
I believe the current code is perfectly fine and if you prefer to keep it like it is (`if it works, don't touch it` :) )I am fine with it. 

However I was going to test the latest code anyway - because I noticed the latest change on the `Concurrent` RS - so I thought some clean-up would benefit future maintenance.

So please let me know if you agree with my changes and I will test it otherwise we can close this PR (by the way, the `A` in my list doesn't really affect performance really much so it's all about simpler code).

luca-s:
Well, I have done a really thorough testing and I do not see any issue with this PR.  

In particular I made sure there are no regressions on all the issues that have been fixed since concurrent was created:
-  premature termination when data still available (#98)
- if a request is empty, rs hangs forever (#98)
- deadlock (#82)

What is missing is to test your fix "[io] Fix concurrent recordstream if close is called from multiple threads", but I don't know how to reproduce it. However I don't believe that is an issue because I didn't change the logic of your fix: multiple concurrent calls to `Concurrent::close()` were dangerous because the mutex was released within `Concurrent::close()`.

gempa-jabe:
I *think* that the balanced RS crashed if the picker window in scolv was closed while data was still streaming. I am currently busy with other stuff, so let me get some time to verify your changes carefully. As I understood there is no pressure on side.

luca-s:
Please take your time and also, if you prefer the current code, feel free to drop this PR. I created it because It helps code maintenance from my point of view, but it is just my opinion and it is not a fundamental change.

luca-s:
@gempa-jabe if you feel like having a look at this PR,  I have been running the code for 2 months and I haven't found any issue.

gempa-jabe:
After checking the changes again, actually calling  `setSource` should restart the stream again. According to the changes, this is not possible anymore and `setSource` would always return `false`. Actually we destroy streams once finished in all SeisComP applications but that is not guaranteed and according to the interface documentation the behaviour is not clearly defined. The semantics w.r.t. this PR is then: once closed it can't be reused ever again. I would really prefer to use it like a file. Open and close as often as you wish. What do you think?

PR #121, opened 2024-01-22
Be a little more tolerant in handling non-perpendicular sensor components

jsaul:
The orientation of three sensor components is supposed to be mutually perpendicular, but this is not always achieved (particularly the separate STS-1 horizontal components). We therefore tolerate up to 5 degrees of deviation from perfect perpendicularity. This includes all deviations from perpendicularity that can be found at least in the global networks.

### DIFF
--- a/libs/seiscomp/datamodel/utils.cpp
+++ b/libs/seiscomp/datamodel/utils.cpp
@@ -382,6 +382,12 @@ bool by_Z_desc_and_code_asc(const ComponentAxis &axis1, const ComponentAxis &axi
 bool getThreeComponents(ThreeComponents &res, const SensorLocation *loc, const char *streamCode, const Core::Time &time) {
 	int len = strlen(streamCode);
 
+	// The orientation of three sensor components is supposed to be
+	// mutually perpendicular, but this is not always achieved (e.g.
+	// sometimes STS-1 horizontal components). We therefore tolerate
+	// up to 5 degrees of deviation from perfect perpendicularity.
+	const float perpendicularityTolerance = 5*M_PI/180;
+
 	res = ThreeComponents();
 
 	std::vector<ComponentAxis> comps;
@@ -437,8 +443,7 @@ bool getThreeComponents(ThreeComponents &res, const SensorLocation *loc, const c
 			for ( size_t a = 0; a < comps.size(); ++a ) {
 				float alpha = comps[a].axis.dot(axis);
 
-				// Allow one degree of uncertainty
-				if ( alpha > 0.0174524064373 ) {
+				if ( alpha > perpendicularityTolerance ) {
 					newOrthogonal = false;
 					break;
 				}


jsaul:
`DataModel::numberOfComponents` is good news. However, in the code block you are probably referring to I am not only counting components, but collecting the matching streams (usually three) for further handling. Therefore it is not a re-implementation of `DataModel::numberOfComponents`.

Regarding discussion with the community. What do you want to discuss? Whether SeisComP should be enabled to process data from certain STS-1's or not?

jsaul:
Note also that in more than 99 percent of the sensors used, the components are already perpendicular by design. The allowance of up to 5 degrees (rather than one degree) simply reflects the reality of some stations in important networks plus some safety margin.

libs/seiscomp/datamodel/utils.cpp
```diff
@@ -379,18 +379,18 @@ bool by_Z_desc_and_code_asc(const ComponentAxis &axis1, const ComponentAxis &axi
...
+	tc = ThreeComponents();
 
+	// First find matching streams based on code and time
+	std::vector<Stream*> matchingStreams;
 	for ( size_t i = 0; i < loc->streamCount(); ++i ) {
 		Stream *stream = loc->stream(i);
 
 		try {
-			if ( stream->end() <= time ) {
+			if ( stream->end() < time ) {
```
gempa-jabe, line 393: I don't think that this is correct. End time is not inclusive: start <= time < end.

jsaul-private: I agree with you.

jsaul: I agree with you.

gempa-dirk:
> The allowance of up to 5 degrees (rather than one degree) simply reflects the reality of some stations in important networks plus some safety margin.

True. So if you think that 5 degrees are acceptable and this is common sense, then go for it.

libs/seiscomp/datamodel/utils.cpp
```diff
@@ -379,13 +379,13 @@ bool by_Z_desc_and_code_asc(const ComponentAxis &axis1, const ComponentAxis &axi
...
+bool getThreeComponents(ThreeComponents &tc, const SensorLocation *loc, const char *streamCode, const Core::Time &time) {
 	int len = strlen(streamCode);
 
-	res = ThreeComponents();
-
-	std::vector<ComponentAxis> comps;
+	tc = ThreeComponents();
 
+	// First find matching streams based on code and time
+	std::vector<Stream*> matchingStreams;
```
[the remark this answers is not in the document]
jsaul: Because I prefer to separate the two steps:

1. collect the matching streams and give up if their number is not 3
2. if 3 streams were found, check if these are perpendicular

In fact this could even be done in separate functions.

To me when I debugged the problem with a specific sensor location, the code was initially not easy to understand and the motivation for the change is to improve this. But YMMV of course.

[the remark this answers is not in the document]
gempa-jabe: As I have showed above, it won't work for all cases. Even if the code might be more easy to understand (which I do not agree with), it removes functionality which I do not prefer. The aim of this PR is to be more tolerant in terms of orientation, but with that change in place we are even less tolerant in some situations and it has actually nothing to do with the topic of this PR. I would like to keep PRs small, simple and focused. Even renaming a variable should be done in a separate commit separated from the actual changes related to this PR.

PR #122, opened 2024-01-23
[filter] Adding recursive envelope filter

jsaul:
Band pass and envelope filter

This is a recursive band pass and envelope filter combination. Strictly speaking the envelope would require frequency-domain filtering because it requires the Hilbert transform of the signal. We cannot compute the _exact_ Hilbert transform using time-domain recursive filters. By applying a rather narrow band pass to the data, however, we can compute an _approximate_ Hilbert transform. This is because for a narrow-band signal the Hilbert transform will be the scaled time derivative. This actually works quite well in practice as long as the band pass is not too wide and has relatively steep corners to block energy from outside the pass band. A band width of one octave and a filter order of four was found to work well in most cases. It is therefore highly recommended to stick to these defaults and only set the center frequency.

This filter may be used in a filter chain. Its name is BPENV and it accepts
up to three arguments:
  - center frequency in Hz (must be specified)
  - bandwidth in octaves (default: 1)
  - filter order (default: 4)

### DIFF
--- a/libs/seiscomp/math/filter/bpenv.cpp
+++ b/libs/seiscomp/math/filter/bpenv.cpp
@@ -1,0 +1,161 @@
+/***************************************************************************
+ * Copyright (C) gempa GmbH                                                *
+ * All rights reserved.                                                    *
+ * Contact: gempa GmbH (seiscomp-dev@gempa.de)                             *
+ *                                                                         *
+ * GNU Affero General Public License Usage                                 *
+ * This file may be used under the terms of the GNU Affero                 *
+ * Public License version 3.0 as published by the Free Software Foundation *
+ * and appearing in the file LICENSE included in the packaging of this     *
+ * file. Please review the following information to ensure the GNU Affero  *
+ * Public License version 3.0 requirements will be met:                    *
+ * https://www.gnu.org/licenses/agpl-3.0.html.                             *
+ *                                                                         *
+ * Other Usage                                                             *
+ * Alternatively, this file may be used in accordance with the terms and   *
+ * conditions contained in a signed written agreement between you and      *
+ * gempa GmbH.                                                             *
+ ***************************************************************************/
+
+
+#include <math.h>
+#include <seiscomp/math/filter/bpenv.h>
+
+namespace Seiscomp {
+namespace Math {
+namespace Filtering {
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+static void checkParameters(double centerFrequency, double bandwidthOctaves, int order) {
+	if ( centerFrequency <= 0.  )
+		throw Core::ValueException("Filter center frequency must be positive");
+	if ( bandwidthOctaves <= 0. )
+		throw Core::ValueException("Filter band width must be positive");
+	if ( order < 0 )
+		throw Core::ValueException("Filter order must be non-negative");
+	if ( order > 10 )
+		throw Core::ValueException("Filter order > 10 makes no sense");
+}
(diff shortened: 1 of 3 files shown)


libs/seiscomp/math/filter/bpenv.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename TYPE>
+BandPassEnvelope<TYPE>::BandPassEnvelope(double centerFrequency, double bandwidthOctaves, int order, double fsamp)
+: _centerFrequency(centerFrequency)
+, _bandwidthOctaves(bandwidthOctaves)
+, _order(order)
+, _samplingFrequency(0)
```
gempa-jabe, line 51: You can with C++11 initialize the variables in the declaration. We think that this is much more readable. You don't need to change that, just a note.

libs/seiscomp/math/filter/bpenv.h
```diff
@@ -0,0 +1,99 @@
...
+
+
+	private:
+		double _centerFrequency;
+		double _bandwidthOctaves;
+		int    _order;
+		double _samplingFrequency;
+		double K, yp;
+
+		IIR::ButterworthBandpass<TYPE> *bp;
```
gempa-jabe, line 87: A SmartPointer might be a better choice to avoid manual deletion.

libs/seiscomp/math/filter/bpenv.cpp
```diff
@@ -0,0 +1,175 @@
...
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename TYPE>
+BandPassEnvelope<TYPE>::~BandPassEnvelope()
+{
+	if (bp)
```
gempa-jabe, line 66: Please add white spaces and curly brackets:

```c++
if ( bp ) {
  delete bp;
}
```
```

File: libs/seiscomp/math/filter/bpenv.h [hunk 4]
```diff
@@ -0,0 +1,99 @@
...
+
+		InPlaceFilter<TYPE> *clone() const override;
+
+
+	private:
+		double _centerFrequency;
+		double _bandwidthOctaves;
+		int    _order;
+		double _samplingFrequency;
+		double K, yp;
```
gempa-jabe, line 85: Use underscores for all class members.

jsaul:
One more remark. I tested this filter extensively and also compared the output with that of a frequency-domain envelope involving an exact Hilbert transform. There is typically some high-frequency noise "riding" on the envelope, due to the imperfection of the recursive approximation of the Hilbert transform. The correlation with the frequency-domain equivalent, however, is usually more than 0.99, which is much better than I had expected.

libs/seiscomp/math/filter/bpenv.h
jsaul: Makes sense

gempa-dirk:
@jsaul When merged, please add the new filter also to the documentation of the filters:
https://github.com/SeisComP/seiscomp/blob/master/doc/base/filter-grammar.rst

jsaul:
Let me know if there is anything that's blocking the merge.

gempa-jabe:
> Let me know if there is anything that's blocking the merge.

Just the holiday season ...

PR #124, opened 2024-02-01
Adding phase count filter and hide f/x/new events checkbuttons

victorsalles:
Added a phase count filter in addition to the existing ones (lat, lon, mag, depth) for the SQL query in scolv. Also added two new check buttons to i) hide final/rejected events and ii) hide new events from the retrieved list.

### DIFF
--- a/libs/seiscomp/gui/datamodel/eventfilterwidget.ui
+++ b/libs/seiscomp/gui/datamodel/eventfilterwidget.ui
@@ -1,10 +1,8 @@
-<ui version="4.0" >
- <author></author>
- <comment></comment>
- <exportmacro></exportmacro>
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
  <class>EventFilter</class>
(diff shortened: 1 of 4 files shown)


gempa-jabe:
@cla-bot recheck

gempa-jabe:
@cla-bot check

gempa-jabe:
Your PR is going to be checked.

gempa-jabe:
Btw, which Qt designer version have you used to modify the ui?

gempa-jabe:
I have checked the code which compiles fine and also works as expected. Thank you.

victorsalles:
> Btw, which Qt designer version have you used to modify the ui?

I've been using Qt Creator 8.0.2. Which version was used to create the ui?

gempa-jabe:
The Qt Creator version is not much important. The Qt version is more of interest. For reasons of compatibility we have been using the designer of Qt 4.2 for the ui. As Qt 4 was set to eol, we are now using the designer of Qt 4.8.7 for maximum portability.

PR #125, opened 2024-02-15
[core] Fix const iterator of IntrusiveList

andres-h:


gempa-jabe:
Thanks, that has been forgotten to be ported from the non-const version.

PR #131, opened 2024-05-21
[stdloc] fix octtree search failures in case of very large cell sizes

luca-s:
@gempa-jabe one last commit to add to the last week series of fixes. I tested `stdloc` at regional scale, comparing the SED `NonLinLoc` and `LOCSAT` solutions to the ones from `stdloc` (in Octree and least-square modes) and, except for this PR bug,  I found the solutions to be consistent between locators. 

That said I do not believe there  is good reason to use `stdloc` over `NonLinLoc` and `LOCSAT`, since they are perfectly fine, but I like the workflow of `scanloc`+`stdloc` because a user can choose the velocity model of their choice and the processing adapts to it. 

Eventually I would like to release a code that enables NLL grids as Seiscomp travel time tables and that would make `scanloc`+`stdloc` workflow interesting at regional scale.

### DIFF
--- a/plugins/locator/stdloc/stdloc.cpp
+++ b/plugins/locator/stdloc/stdloc.cpp
@@ -1194,9 +1194,13 @@ void StdLoc::computeProbDensity(const PickList &pickList,
 	}
 
 	rms = 0.0;
+	double sigma = _currentProfile.gridSearch.travelTimeError;
 
 	double l1SumWeightedResiduals = 0.0;
 	double l2SumWeightedResiduals = 0.0;
+	double weightedSigma = 0.0;
+	double weightedSquaredSigma = 0.0;
+	double sumWeights = 0.0;
 	double sumSquaredWeights = 0.0;
 
 	for ( size_t i = 0; i < pickList.size(); ++i ) {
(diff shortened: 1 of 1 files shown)


gempa-jabe:
Actually I like the idea of a well maintained and locator using modern C++ standards. There are probably features in LOCSAT which stdloc does not cover like backazimuth/slowness utilization and station corrections. How do the locators compare performance-wise, have you tested that?

I would like to get rid of the old locator code that 1. nobody understands and 2. nobody maintains anymore. Not very realistic though ...

luca-s:
> There are probably features in LOCSAT which stdloc does not cover like backazimuth/slowness utilization and station corrections.

That would be easy to add, but I am more worried about teleseismic events. I have never tried stdloc with those. 

> How do the locators compare performance-wise, have you tested that?

I didn't run a comparison, but speed-wise using stdloc in iteretive least-square mode is very fast. In the experiments I work with we have ~100 events per minute and we use scanloc+stdloc without delays.

> Not very realistic though ...

Agree, it takes time to achieve that. It is not about being difficult to add features to stdloc, it is about testing and testing and testing

PR #132, opened 2024-05-21
Update global_gui.xml

comoglu:
The information in the description is misleading. It requires that projection names be "Rectangular" and "Mercator," otherwise it will throw an error as "Projection rectangular not available, defaulting to Rectangular"


gempa-jabe:
Thanks.

PR #135, opened 2024-08-03
Improvements to Flinn-Engdahl region names

nackerley:
The important thing here is that Senegal is spelled "Sengal", which is incorrect.

Additionally, since I could find no other French translation elsewhere (https://en.wikipedia.org/wiki/Flinn-Engdahl_regionalization only has translations to Spanish and German the same languages other than English supported by https://github.com/jsaul/flinnengdahl), I did my own translation to French. In the process, I found that the (inconsistent!) use of abbreviations really got in the way, so here I've also eliminated the abbreviations.

If there is a reason to keep the abbreviations, let me know and I'll initiate a pull request that only corrects the spelling of Senegal.

### DIFF
--- a/libs/seiscomp/seismology/regions/fedata/names.asc
+++ b/libs/seiscomp/seismology/regions/fedata/names.asc
@@ -31,8 +31,8 @@ Off Coast of Oregon
 Near Coast of Oregon
 Oregon
 Western Idaho
-Off Coast of N. California
-Near Coast of N. California
+Off Coast of Northern California
+Near Coast of Northern California
 Northern California
 Nevada
 Off Coast of California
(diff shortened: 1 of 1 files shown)


jsaul:
The names were inherited from old Fortran codes and at some point turned into a more readable form as the original names where all upper case... The abbreviations were probably due to tight space constraints in the Fortran codes.

jsaul:
ObsPy actually maintains the original list of names here: https://github.com/obspy/obspy/blob/master/obspy/geodetics/data/names.asc?plain=1
"Papua New Guinea" is fully spelled out where space permits, and abbreviated as "P.N.G." and even "PNG.".
And "SENGAL" is there, too. ;)

gempa-jabe:
I personally would prefer Wikipedia as source for the names but that is just my opinion. Obspy is still using the very old version with IRIAN JAYA, so they might have taken that from the same source as we did 18 years ago ;)

In the end  I do not want to offend anyone with old region names. Even putting only the FE region ID would be sufficient and anyone can grep her own translation.

gempa-jabe:
Don't worry, I just wanted to add my opinion w.r.t. this PR and I understood your comment as how you intended it.

PR #136, opened 2024-08-06
[ttt/homogeneous] Try to read application configuration first

gempa-jabe:
Currently the homogeneous was reading its parameters exclusively from global.cfg at all three configuration stages. This makes it impossible to configure it per application, e.g. in app.cfg.

### DIFF
--- a/libs/seiscomp/seismology/ttt/homogeneous.cpp
+++ b/libs/seiscomp/seismology/ttt/homogeneous.cpp
@@ -19,6 +19,7 @@
 
 
 
+#include <seiscomp/system/application.h>
 #include <seiscomp/seismology/ttt.h>
 #include <seiscomp/core/strings.h>
 #include <seiscomp/math/geo.h>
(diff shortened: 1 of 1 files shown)


gempa-jabe:
@luca-s please have a look. I came across this issue by testing an application and defining the parameters in the `app.cfg`. It did not work. It now tries to get the global application parameters in memory first and if that does not work (because there is no app) it will read them from `global.cfg` as previously. Any objections from your side?

luca-s:
Thank you for looking into this @gempa-jabe . The code was written in that way because I didn't know the correct way to keep into account all the configuration stages. I trust your knowledge so no objections from me.

gempa-jabe:
OK, thank you.

FYI: `Environment::initConfig(...)` reads all configuration files and populates the `Config` object. As the provided name is empty, it would try to read, e.g. `seiscomp/etc/.cfg`, which does not exist. In order to get the current application name (the environment has no idea of an application), you would need to get the global application instance and ask for it or to also support command line override (such as `--ttt.homogeneous.test.P-velocity=x`) get the global application configuration object and use it. A shortcoming of the TTT interface is that it does not has an `init` method like the locator which gets the config object passed as parameter.

luca-s:
Very useful information thanks.

>  A shortcoming of the TTT interface is that it does not has an init method like the locator which gets the config object passed as parameter.

How does the code that calls the `LocatorInterface::init(Config)` method retrieve the `Config` object? The same as your code in this PR?

gempa-jabe:
`SCApp->configuration()` which retrieves the app instance. Without a GUI, one can also use `SCScoreApp->configuration()`. See `originlocatorview.cpp` for example.

PR #139, opened 2024-08-22
[stdloc] set enableConfidenceEllipsoid to false by default

luca-s:
Despite my efforts to fix the eigv/chi2 code, I still see random crashes (it happened only one, but I was not able to reproduce it). That library is a nightmare so I do not trust that code very much and I would like to disable the computation of the confidence ellipsoid by default.

The common usage of `stdloc` is to be used by `scanloc` and in this case the confidence ellipsoid is not required.

If that code crashes when called from scolv that is annoying, but the user notices the issue. However, If the code crashes when called by scanloc than who is going to notice that?

### DIFF
--- a/plugins/locator/stdloc/descriptions/global_stdloc.xml
+++ b/plugins/locator/stdloc/descriptions/global_stdloc.xml
@@ -61,7 +61,7 @@
 							Confidence level, between 0.5 and 1.0, used in computing the confidence ellipsoid.
 							</description>
 						</parameter>
-						<parameter name="enableConfidenceEllipsoid" type="boolean" default="true">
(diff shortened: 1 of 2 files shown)


gempa-dirk:
> If that code crashes when called from scolv that is annoying, but the user notices the issue. However, If the code crashes when called by scanloc than who is going to notice that?

In this case, the application also crashes and you may restart using seiscomp check or monitor core dumps.

gempa-jabe:
This code needs an overhaul (I agree with your opinion). Random crashes indicate uninitialized variables. Test cases which reproduce crashes would be nice to include them into the automatic tests.

PR #140, opened 2024-08-29
[stdloc] update documentation

luca-s:
Better explain the use case for each location method and add an example configuration as well.

### DIFF
--- a/plugins/locator/stdloc/descriptions/global_stdloc.rst
+++ b/plugins/locator/stdloc/descriptions/global_stdloc.rst
@@ -1,6 +1,6 @@
 StdLoc is a SeisComP locator plugin that combines standard location methods
 and was developed with the focus on local seismicity, although the methods
-are generic enough to work at different scales.
+are generic enough to work at larger scales as well.
 
 Plugin
 ======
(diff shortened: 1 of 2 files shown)


plugins/locator/stdloc/descriptions/global_stdloc.rst
```diff
@@ -12,47 +12,95 @@ How does it work?
...
+  location estimate and the location uncertainty is completely known. However the
+  method is very slow. It can be used to relocate events in :ref:`scolv` that seem
+  difficult to locate via other methods or to verify the uncertainty of a solution.
+  The following example configuration computes a grid search around the average
+  location of the picked stations. The grid points are spaced apart 0.5km
+  horizontally and 2km vertically.
+
+   .. code-block:: params
+
+      method = LeastSquares
```
gempa-dirk, line 41: Should this not be _GridSearch_ ?

gempa-dirk (changes requested):

plugins/locator/stdloc/descriptions/global_stdloc.rst
luca-s: Yes, thank you for spotting it.

PR #146, opened 2024-10-17
[magnitudes] Request for adding IGN magnitudes

Elodoin:
Dear developers, 

We would like to add the amplitude and magnitude processors of the magnitudes used in routine analysis at the National Geographic Institute in Spain to SeisComp. 

We have derived the amplitude and magnitude processors from the existing ones. We have not used magnitude-aliases, because we needed a further configuration. Each magnitude have been referenced properly in their corresponding files. 

We don't know if we should make the pull request in this way or we have to send it by another way.

Thank you in advance!

Kind regards,

Jaime Barco and Eduardo Suarez


jsaul:
Thanks a lot also from my side. This is a great contribution!

Regarding the integration I agree with @gempa-jabe that it would be ideal to have the IGN as an "addon" package. This makes it easier for novice users to first concentrate on the "core" packages, while more specialized packages such as regional magnitudes are anyway more likely to target advanced users.

There are good examples for contributed packages, e.g. [contrib-gns](https://github.com/SeisComP/contrib-gns) or [contrib-ipgp](https://github.com/SeisComP/contrib-ipgp). As GFZ we have also contributed some addons like the scdlpicker package or scmert. An additional advantage of a separate package (under the SeisComP umbrella) is enhanced visibility of your code contribution.

gempa-dirk:
In MLc we are trying to incorporate configurable features which cover a wide range of amplitude and magnitude requirements and we are happy to add more feature which are currently missing for your applications. We have started this approach when we realized that the number of magnitude types was growing and started to be confusing to operators.

It may the case that your requirements are identical or similar to those also required by others. So why not making them available to the community? Since your code for MLcan is very similar to MLc adding your requirements may be not be a big problem. Can you point out the amplitude specifications/difference or add your missing code the existing one for MLc and provide the original scientific references?

Elodoin:
Thanks all for your clarifications!

We will follow your suggestions and, probably, implement it as a contributed packages the Veith-Clawson 1972 magnitude at least. The difference between mb_VC and the current mb, are the [Q-factors](https://github.com/Elodoin/sc_common_magnitudes/blob/master/libs/seiscomp/seismology/mb_VC.cpp), which we think that implement them as a variable on the rst files is a bit messy. 

For the mb_Lg we will try to use the "alias" as a derivative from the MLc. 

As @gempa-dirk points out, [MLcan](https://github.com/Elodoin/sc_common_magnitudes/blob/master/libs/seiscomp/processing/magnitudes/MLcan.cpp) is quite similar to MLc. The main difference is that we aim to implement a station-specific magnitude correction factor per channel, as calculated by [Mezcua & Rueda (2022)](https://link.springer.com/article/10.1007/s00445-022-01553-9) for the Canary Islands. We are unsure whether it would be better to submit a pull request to MLc to incorporate this channel-based correction, or to leave it as part of the contributed package.

Again, thanks for your time and your kind words!

Issue #154, opened 2025-01-13
[libtau + LOCSAT ttt] throw NoPhaseError when source depth is negative


luca-s:
Hi @gempa-jabe ,

as discussed in person, I believe that `LOCSAT` and `libtau` should throw a `NoPhaseError` when queried with a negative source depth, since they cannot return correct data. In the GUI it might not be a big issue, but it is annoying to see code such as `stdloc` to fail because the ttt returns wrong data.  

You can quickly verify that the both `LOCSAT` and `libtau` do not properly work with negative source depth by printing the travel time tables at constant distance and varying depth:

```
from seiscomp.seismology import TravelTimeTableInterface

locsat = TravelTimeTableInterface.Create("LOCSAT")
locsat.setModel("iasp91_scanloc")  # you can also use iasp91

libtau = TravelTimeTableInterface.Create("libtau")
libtau.setModel("iasp91")

def p(ttt, depth):
  tdata = ttt.compute("P", 0.1, 0.1, depth, 0., 0., 0)
  print("depth %.1f km: time %.2f takeoff %.1f dtdd %.3f dtdh %.4f" %
          (depth, tdata.time, tdata.takeoff, tdata.dtdd, tdata. dtdh))
```

```
print("LOCSAT")
for depth in range(5, -6, -1):
    p(locsat, depth)
```

```
LOCSAT
depth 5.0 km: time 2.84 takeoff 108.2 dtdd 18.176 dtdh 0.0537
depth 4.0 km: time 2.79 takeoff 104.2 dtdd 18.505 dtdh 0.0421
depth 3.0 km: time 2.76 takeoff 100.9 dtdd 18.783 dtdh 0.0324
depth 2.0 km: time 2.73 takeoff 94.7 dtdd 18.957 dtdh 0.0139
depth 1.0 km: time 2.72 takeoff 94.6 dtdd 19.063 dtdh 0.0139
depth 0.0 km: time 2.70 takeoff 94.6 dtdd 19.170 dtdh 0.0139
depth -1.0 km: time 2.69 takeoff 94.6 dtdd 19.277 dtdh 0.0139
depth -2.0 km: time 2.67 takeoff 94.6 dtdd 19.383 dtdh 0.0139
depth -3.0 km: time 2.66 takeoff 94.5 dtdd 19.490 dtdh 0.0139
depth -4.0 km: time 2.65 takeoff 94.5 dtdd 19.596 dtdh 0.0139
[...]

PR #155, opened 2025-01-15
[homogeneous] throw out_of_range instead of NoPhaseError

luca-s:
Use std::out_of_range exception when queried for data outside the user defined model ranges.


luca-s:
See #154

PR #156, opened 2025-01-15
[stdloc] add handling of IgnoreInitialLocation capability

luca-s:


luca-s:
This is just a minor feature. I was playing with `scolv` and `stdloc` and noticed this option was useful, but missing.

gempa-jabe:
There is nothing blocked. I can easily merge it.

PR #162, opened 2025-06-17
fix compile issue against latest libxml2

Donavin97:
This PR serves to fix an error when compiling against the latest libxml2 > 2.14.

### DIFF
--- a/libs/seiscomp/io/archive/xmlarchive.cpp
+++ b/libs/seiscomp/io/archive/xmlarchive.cpp
@@ -484,7 +484,8 @@ void XMLArchive::close() {
 
 	_forceWriteVersion = -1;
 
-	initGenericErrorDefaultFunc(nullptr);
+        xmlStructuredErrorFunc handler = xmlStructuredErrorHandler;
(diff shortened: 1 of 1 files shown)


gempa-jabe:
Thank you for the fix. I think that we have fixed that already with b6adf1c06a86c4ff3931af3421c9edec882e6919. Please try the latest master branch on your end. Your fork is a little bit behind the latest code.

Donavin97:
Thanks for clarifying. Will have a look.

Donavin97:
Good morning,
I added a new lable to pickerview.ui and pickerview.cpp.
AmpLabel reads out amplitudes when scroling through the traces.
It uses raw values though.
I would see if adding an invisible progress bar might make the traces audable.
Orca, The screen reader I use, can output progress bars as beeps.
The higher the progress bar, the higher the beep.
So, I have a working implimentation now.
I tested it, and I could make fairly accurate picks, I just have to zoom in quite a bit using control + right.
Thanks.

Donavin97:
Hi @jabe
The audable picker is complete.
I just want you to test if the progress bar interfeers with the normal workings of scolv.
Thanks.

PR #187, opened 2026-03-31
filterpicker implementation

Donavin97:
- **(PLUGIN) Added filterpicker based on lomax et al. (2012**
- **(FILTER) Filterpicker inplaceFilter implementation for use in filter chain**


gempa-lukas:
Hey @Donavin97,

I have tested the picker and the filter and have some comments.

The filter itself looks good to me. Based on my short testing, I think this can be a good alternative to the STA/LTA filter as detecFilter.

For the picker, I was not that satisfied with my test results. Maybe I just miss configured it or made some mistakes. 
Therefore, do you have some test examples and configuration/bindings? Like mseed and event.xmls with the configuration/bindings would help to verify.
Or do you even have evaluation results that show somehow the capability of the picker? 


In addition, I think some parts of the code need a review. Exemplary:
- name of  "descriptions/global_filterpicker..xml" -> double dots
- in the processing/filterpicker/CMakeLists.txt the lines need to be added
`FILE(GLOB descs "${CMAKE_CURRENT_SOURCE_DIR}/descriptions/*.xml")`
`INSTALL(FILES ${descs} DESTINATION ${SC3_PACKAGE_APP_DESC_DIR})`
- I suggest to add more configuration parameters, like to enable/disable the polarity picker (taking the polarity value)
- And there was a bug in the processing/filterpicker/filterpicker.cpp code, in the FilterPicker::initFilterBank() function , that prevented the correct usage of the configured minFreq in offline playbacks, as initially  _stream.fsamp is empty and therefore the defaults will be taken

Kind regards,
Lukas

Donavin97:
Hi @gempa-lukas.
Thanks for checking it out.
I have worked on the implementation a bit and fixed the bug where stream.fsamp could be 0.
I also fixed the description name and added the correct install code in CMake_lists.txt.
I will further add some more config parameters as you suggested.
I can send a small example of data if you would like.
I have a very limited dataset to work with though, but I think it might help.
The reason for the developement was to detect very week signals with a broad set of filters, meaning suttle events can be picked up.
For example, mining events, and tectonic events all in one filter configuration.
If you need anything else, feel free to reach out.
Best:
Donavin.

PR #189, opened 2026-04-13
[common] Extend City<T> with country, AdminRegion and type; enrich cities.xml

comoglu:
Extends `City<T>` with three new fields to enable richer location descriptions in SeisComP applications.

### `AdminRegion` struct (replaces flat `state`/`stateFull` strings)

A dedicated struct serialised as a child element:

```xml
<state abbr="NSW"><name>New South Wales</name></state>
```

- `abbr` — ISO 3166-2 alphabetic subdivision suffix (e.g. `NSW`, `CA`); empty where only numeric codes exist (e.g. Japan, China, Türkiye)
- `name` — full subdivision name

### `country` field

Full English country name serialised as a `<country>` child element:

```xml
<country>Australia</country>
```

Allows applications to display a human-readable country name without maintaining a separate ISO A2 → name lookup table.

### `type` field

Location type derived from GeoNames feature codes. One of (or empty if unknown):

| Value | Meaning |
|-------|---------|
| `city` | Capital or administrative centre (PPLC, PPLA, PPLA2) |
| `town` | Populated place or minor admin centre (PPL, PPLA3, PPLA4) |
| `village` | Small settlement (PPLF, PPLL, PPLR, PPLS, etc.) |
| `suburb` | Section of a populated place (PPLX) |

## Updated `cities.xml`

Re-enriched from freely available sources:

| Field | Source | Coverage |
|-------|--------|----------|
| `countryID` | Natural Earth 10m country polygons | 71,682 / 71,685 |
| `country` | Natural Earth SQLite (`ne_10m_admin_0_countries`) | 70,242 |
| `state` | Natural Earth `ne_10m_admin_1_states_provinces` (ISO 3166-2 abbreviations) | 71,654 |
| `type` | GeoNames `cities500.txt` feature codes | 56,860 |

All fields are optional and additive — existing applications ignoring them are unaffected.

## Enrichment tooling

The pipeline script used to produce the enriched `cities.xml` is available at
https://github.com/comoglu/cities-xml-update — allowing operators to add their own cities or re-enrich from updated upstream data.

### DIFF
--- a/libs/seiscomp/math/coord.cpp
+++ b/libs/seiscomp/math/coord.cpp
@@ -283,7 +283,7 @@ const std::string &City<T>::category() const {
 
 // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 template<typename T>
-void City<T>::setType(const std::string &t) {
+void City<T>::setType(CityType t) {
 	_type = t;
 }
 // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
@@ -293,7 +293,7 @@ void City<T>::setType(const std::string &t) {
 
 // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 template<typename T>
-const std::string &City<T>::type() const {
+CityType City<T>::type() const {
 	return _type;
 }
 // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
--- a/libs/seiscomp/math/coord.cpp
+++ b/libs/seiscomp/math/coord.cpp
@@ -251,10 +251,73 @@ const std::string &City<T>::category() const {
 
 // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 template<typename T>
+void City<T>::setType(const std::string &t) {
+	_type = t;
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename T>
+const std::string &City<T>::type() const {
+	return _type;
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename T>
+void City<T>::setState(const std::string &s) {
+	_state = s;
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename T>
+const std::string &City<T>::state() const {
+	return _state;
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+template<typename T>
+void City<T>::setStateFull(const std::string &s) {
+	_stateFull = s;
+}
+// <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+
+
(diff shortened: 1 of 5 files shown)


libs/seiscomp/math/coord.h
```diff
@@ -104,12 +104,27 @@ class City : public NamedCoord<T> {
...
+		void setStateFull(const std::string &);
+		const std::string &stateFull() const;
+
 		void serialize(Core::BaseObject::Archive& ar) override;
 
 	private:
 		std::string _countryID;
 		double _population;
 		std::string _category;
+		std::string _type;
```
gempa-jabe: Wouldn't a type enumeration be a better approach? Is there a defined set of values for type?

libs/seiscomp/math/coord.h
```diff
@@ -104,12 +104,27 @@ class City : public NamedCoord<T> {
...
+		std::string _state;
```
gempa-jabe: I think a dedicated class with an abbreviation would be more appropriate. In XML something like:

```xml
<state short|abbr|younameit="BRB">Brandenburg</state>
```

gempa-jabe:
I would be interested in how big that will grow. Given all the redundancy of state and type and also country cannot be for free. I haven't checked but maybe there is already a standard out there to describe such administrative information.

comoglu:
[cities_enriched.xml](https://github.com/user-attachments/files/26674026/cities_enriched.xml)

comoglu:
15MB

gempa-jabe:
OK, that is not too much of a difference. The schema itself would be something I would like to address. I personally am not a big fan of "stateFull" which is actually "state". Unfortunately I do not have the time right now.

comoglu:
Thanks for the feedback. To address the stateFull naming concern: in the updated implementation I have already replaced both state and stateFull flat strings with a dedicated AdminRegion that has two clear fields — abbr (ISO 3166-2 suffix, like. "NSW") and name (full subdivision name, e.g. "New South Wales"). This serializes as a <state> child element:

<state abbr="NSW"><name>New South Wales</name></state>

comoglu:
Updating Cities xml for entire world in one rule is indeed a hard work.

comoglu:
[cities_enriched.xml](https://github.com/user-attachments/files/26674610/cities_enriched.xml)

comoglu:
I've shared the enriched cities.xml as a preview. Enriching city types globally with a single rule set is harder than I would expect — GeoNames feature codes are inconsistently applied (e.g. a 1M+ population city classified as PPL → "town" based on natural earth data (shape file)).

Before going further, I'd appreciate your direction:

Is the type field worth pursuing, or should I drop it ?
For state: the ISO 3166-2 abbreviation approach works well for ~40 countries (US, AU, CA, IN) but many countries only have numeric codes. I would go with partial coverage, if it is OK, or would you prefer I omit state for countries without alphabetic abbreviations?

libs/seiscomp/math/coord.h
comoglu: Good point @gempa-jabe the values are indeed fixed. GeoNames feature codes map to exactly four types: city (PPLC/PPLA/PPLA2), town (PPL/PPLA3/PPLA4), village (PPLF/PPLL/PPLR etc.) and suburb (PPLX), plus Unknown for entries where the attribute is absent.

I've replaced the std::string with a CityType enum class and added parseCityType()/toString() helpers for the XML round-trip. The cities.xml format is unchanged — serialization still uses the lowercase string attribute, so there's no migration needed
https://github.com/comoglu/common/commit/2b7623a1ce6961b395ea26721476d45c4a8a05c4

gempa-jabe:
Actually I do not have strong feelings about it. Maybe we can start extending the city model with the new attributes. You can then test and optimize your `cities.xml` locally. Updating the `cities.xml` can be done in a separate PR. What do you think?

libs/seiscomp/math/coord.cpp
```diff
@@ -83,6 +83,18 @@ void Coord<T>::serialize(Archive& ar) {
 }
 // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
+// AdminRegion implementation
+
+
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+void AdminRegion::serialize(Core::BaseObject::Archive& ar) {
+	abbr = ""; name = "";
```
gempa-jabe: If you would write an object, it would reset your attributes. Maybe checking `isReading()` could help in that regard. Why would you want to do  that anyway? Is there any use-case you have in mind requiring this reset?

comoglu: You're right, I just followed the same pattern already in the file. Since the strings default to empty I'll remove it. Should the pre-existing resets in City<T>::serialize() be cleaned up as well or leave them as-is?

gempa-jabe: I will remove them in the main branch as this is just wrong. Since we are just reading those classes, it might not be important but it is still wrong.

comoglu:
> Actually I do not have strong feelings about it. Maybe we can start extending the city model with the new attributes. You can then test and optimize your `cities.xml` locally. Updating the `cities.xml` can be done in a separate PR. What do you think?

I agree that makes sense. The model extensions (country, AdminRegion and CityType) are already in this PR. I can split the updated cities.xml into a separate PR. Do you think then this PR can be merged after you finish reviewing it. That also gives me time to test and optimise the data locally before submitting it.

For future attributes, elevation could be a natural addition to the dataset. I'll think about what else might be useful during local testing.

gempa-jabe:
> Do you think then this PR can be merged after you finish reviewing it.

Yep, this is actually just an extension to the current model.

libs/seiscomp/math/coord.cpp
comoglu: Done, removed in 5f11d4a3.

comoglu:
> > Do you think then this PR can be merged after you finish reviewing it.
> 
> Yep, this is actually just an extension to the current model.

Great, thanks.

libs/seiscomp/math/coord.cpp
```diff
@@ -327,7 +327,7 @@ const AdminRegion &City<T>::adminRegion() const {
 template<typename T>
 void City<T>::serialize(Core::BaseObject::Archive& ar) {
 	NamedCoord<T>::serialize(ar);
-	_category = ""; _countryID = ""; _country = ""; _type = "";
+	_category = ""; _countryID = ""; _country = ""; _type = CITYTYPE_UNKNOWN;
```
gempa-jabe: Please remove that as well.

gempa-jabe: I have removed all other occurrences with e3a89244cac06a3611c5d81e7da6ff9c612c69e7.

comoglu:
Rebased onto current main to incorporate upstream commit `e3a89244` ([math] Remove variable resets in serialize) and 23 other upstream commits. Conflicts in `coord.cpp` serialize methods were resolved by dropping the reset lines (aligned with the upstream change). Functional diff is unchanged.

gempa-stephan:
One minor change request: Could please use [American English](https://docs.gempa.de/seiscomp/current/base/style-guide.html#english-language) in your source code as well as in the package documentation and description.xml files.

comoglu:
Done

PR #191, opened 2026-04-14
[datamodel] Add generic event filter expression infrastructure

comoglu:
## What this does

Adds a small filter library to `seiscomp/datamodel` that lets any application evaluate a condition string against an event and its preferred origin or magnitude.

The condition syntax builds on the existing `leparser` in `seiscomp/utils` for boolean logic (`&` AND, `|` OR, `!` NOT, parentheses for grouping). Leaf conditions take the form `key==value`, `key<value`, `key>=value` etc. Because the leparser treats `!` as the logical NOT operator, not-equal is written as `!(key==value)` rather than `key!=value`.

## New API

Three classes in `seiscomp/datamodel/filter.h`:

- **`FilterContext`** — abstract interface; subclasses provide named key/value lookup for a specific data object
- **`FilterExpression`** — abstract expression node; trees are built by `parseFilterExpression()`  
- **`EventFilterContext`** — concrete binding of `Event` + `Origin` + `Magnitude` to the filter interface

Free function:

```cpp
FilterExpression *parseFilterExpression(const std::string &condition,
                                        std::string &errorMsg);
```

## Supported keys

String keys: `type`, `evaluationstatus`, `evaluationmode`, `typecertainty`, `agencyid`, `author`

Numeric keys: `lat`, `lon`, `depth`, `magnitude`, `rms`, `azimuthalgap`

String values are normalised to lowercase with spaces and underscores converted to hyphens on both sides of a comparison, so `"not existing"` and `"not-existing"` match the same enum string.

## Example

```cpp
std::string err;
std::unique_ptr<FilterExpression> expr(
    parseFilterExpression("type==earthquake & evaluationstatus==preliminary", err));

EventFilterContext ctx(event, origin, magnitude);
if ( expr && expr->eval(ctx) ) {
    // matched
}
```

## leparser.h change

`_lhs`/`_rhs` in `BinaryOperator` and `_rhs` in `UnaryOperator` are changed from `private` to `protected` so that subclasses can implement `eval()`.

## Notes

This PR is the dependency for a companion PR in `seiscomp/main` that uses `EventFilterContext` to drive configurable event list row highlighting in scolv.

### DIFF
--- a/libs/seiscomp/datamodel/CMakeLists.txt
+++ b/libs/seiscomp/datamodel/CMakeLists.txt
@@ -15,6 +15,7 @@ SET(DM_SOURCES
 	publicobject.cpp
 	diff.cpp
 	utils.cpp
+	filter.cpp
 )
 
 SET(DM_HEADERS
@@ -27,6 +28,7 @@ SET(DM_HEADERS
 	publicobject.h
 	diff.h
 	utils.h
+	filter.h
 	${CORE_DATAMODEL_GENERATED_HEADERS}
 )
 
--- a/libs/seiscomp/datamodel/filter.cpp
+++ b/libs/seiscomp/datamodel/filter.cpp
@@ -1,0 +1,349 @@
+/***************************************************************************
+ * Copyright (C) gempa GmbH                                                *
+ * All rights reserved.                                                    *
+ * Contact: gempa GmbH (seiscomp-dev@gempa.de)                             *
+ *                                                                         *
+ * GNU Affero General Public License Usage                                 *
+ * This file may be used under the terms of the GNU Affero                 *
+ * Public License version 3.0 as published by the Free Software Foundation *
+ * and appearing in the file LICENSE included in the packaging of this     *
+ * file. Please review the following information to ensure the GNU Affero  *
+ * Public License version 3.0 requirements will be met:                    *
+ * https://www.gnu.org/licenses/agpl-3.0.html.                             *
+ *                                                                         *
+ * Other Usage                                                             *
+ * Alternatively, this file may be used in accordance with the terms and   *
+ * conditions contained in a signed written agreement between you and      *
+ * gempa GmbH.                                                             *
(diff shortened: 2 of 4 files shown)


comoglu:
@gempa-jabe  any comment on this?

gempa-jabe:
I haven't found the time yet to check it but I will do.

gempa-jabe:
I took a look and after doing that I have reworked the current `LeParser` which was much overcomplicated IMHO. The new API is easier to digest and it supports the `eval` method with a context (`const void *`) parameter. Furthermore the keywords can be customized to support `&&` as and operator which I prefer over `&`. Also special keywords are supported which are ignored. The filter implementation could add `!=` to support it straight in the comparison. I would keep the generic key-value operation parser and implement a specialized version for the event list.

gempa-jabe:
That required also changed in `scimex` and I hope that I have not broken it. It must be tested.

comoglu:
Are you planning to push your changes? So I can update filter.cpp.

gempa-jabe:
I have uploaded the changes. Actually there is currently no need to define an additional filter in `utils/` itself. The important class to check is now `LeKeyValueContext` which resolves all the values. I added a [test](https://github.com/SeisComP/common/blob/main/libs/seiscomp/test/utils/leparser.cpp) which demonstrates almost all the different options. Tests are still missing for invalid expression strings and error handling.

Implementing a specialized `LeKeyValueContext` especially for the `EventListView` in its private namespace should be possible now to make your configuration work. Error handling has been replaced by exceptions rather than error flag and string access methods.

comoglu:
@gempa-jabe Closing this PR. Your new LeKeyValueContext / LeKeyValueFactory API in leparser.h does exactly what I needed from filter.h, and in a more reusable way. I've migrated the scolv event-list highlight feature (main#115) to use it directly in an anonymous-namespace EventKeyValueContext, so no new shared headers are needed. Thanks for the clean design as always.

PR #193, opened 2026-04-16
[seismology] Add computeAzimuthalGaps() and fix secondary gap in compile()

comoglu:
## Problem

`populateQuality()` / `compile()` in `seismology/locator/utils` computes the primary azimuthal gap but never sets `secondaryAzimuthalGap` in `OriginQuality`, leaving the field unset for any locator that relies on this utility (locsat, etc.).

Additionally, the sort-and-scan logic for computing azimuthal gaps is duplicated across scautoloc (`determineAzimuthalGaps` / `geoProperties`), stdloc, and scrtdd — each with a slightly different implementation.

## Fix

- Add a standalone library function `computeAzimuthalGaps(const std::vector<double> &azimuths, double *primary, double *secondary)` to `seismology/locator/utils` that accepts raw azimuths and computes both the primary and secondary gaps.
- Update `compile()` to call `computeAzimuthalGaps()` and populate `secondaryAzimuthalGap`.

This gives other modules (scautoloc, future locators) a single authoritative implementation to call, as discussed with @jsaul.

## Related

Companion PR in SeisComP/main: scautoloc sc3adapters — export `secondaryAzimuthalGap` on the converted origin.

### DIFF
--- a/libs/seiscomp/seismology/locator/utils.cpp
+++ b/libs/seiscomp/seismology/locator/utils.cpp
@@ -30,6 +30,35 @@
 namespace Seiscomp{
 
 
+bool computeAzimuthalGaps(const std::vector<double> &azimuths,
+                          double &primary, double &secondary) {
+	if ( azimuths.size() < 2 ) {
+		return false;
+	}
+
+	std::vector<double> azi(azimuths);
+	std::sort(azi.begin(), azi.end());
+
+	primary = secondary = 0.;
+	size_t n = azi.size();
+	azi.push_back(azi[0] + 360.);
+	azi.push_back(azi[1] + 360.);
+
+	for ( size_t i = 0; i < n; ++i ) {
+		double g1 = azi[i + 1] - azi[i];
+		if ( g1 > primary ) {
+			primary = g1;
+		}
+		double g2 = azi[i + 2] - azi[i];
+		if ( g2 > secondary ) {
+			secondary = g2;
+		}
+	}
+
+	return true;
+}
+
+
 void compile(DataModel::OriginQuality &quality, const DataModel::Origin *origin) {
 	int usedPhases = 0;
 	int usedDepthPhases = 0;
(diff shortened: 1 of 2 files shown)


libs/seiscomp/seismology/locator/utils.cpp
```diff
@@ -30,6 +30,30 @@
 namespace Seiscomp{
 
 
+bool computeAzimuthalGaps(const std::vector<double> &azimuths,
+                          double *primary, double *secondary) {
+	if ( azimuths.size() < 2 )
```
gempa-jabe: Please add curly brackets to the block even if it is just a one-liner.

libs/seiscomp/seismology/locator/utils.cpp
```diff
@@ -30,6 +30,30 @@
...
+	std::vector<double> azi(azimuths);
+	std::sort(azi.begin(), azi.end());
+
+	*primary = *secondary = 0.;
+	size_t n = azi.size();
+	azi.push_back(azi[0] + 360.);
+	azi.push_back(azi[1] + 360.);
+
+	for ( size_t i = 0; i < n; ++i ) {
+		double g1 = azi[i+1] - azi[i];
```
gempa-jabe: Please use spaces around the arithmetic operator: `i + 1`

gempa-jabe (changes requested):
Thank you, it is looking good to me.

jsaul:
Looks good! Personally I would now prefer to pass `primary` and `secondary` as references rather than pointers.

comoglu:
> Looks good! Personally I would now prefer to pass `primary` and `secondary` as references rather than pointers.

Done.

libs/seiscomp/seismology/locator/utils.cpp
```diff
@@ -30,6 +30,31 @@
...
+	std::sort(azi.begin(), azi.end());
+
+	primary = secondary = 0.;
+	size_t n = azi.size();
+	azi.push_back(azi[0] + 360.);
+	azi.push_back(azi[1] + 360.);
+
+	for ( size_t i = 0; i < n; ++i ) {
+		double g1 = azi[i + 1] - azi[i];
+		if ( g1 > primary ) primary = g1;
```
gempa-jabe: We want brackets around any if-block, regardless whether it is just one line. If-one-liners are not preferable for better readability and error prevention.

gempa-jabe: I will merge it later and make the change unless you want to do it right now.

comoglu: It should be OK now.

PR #195, opened 2026-04-24
[gui] Move event list row-highlight rules into EventListView

comoglu:
## What this does

Moves the configurable row-highlight feature into `EventListView` so all GUI tools using the widget (scolv, scesv, scmvx, …) benefit automatically — verified working in all three.

## How it works

`EventListView` reads `eventlist.highlight.*` rules at construction time using `Utils::LeKeyValueContext` / `LeKeyValueFactory` from `seiscomp/utils/leparser.h`. Highlight is applied in `addEvent()` and re-evaluated in `notifierAvailable()` when the preferred origin or event is updated.

An `EventKeyValueContext` (anonymous namespace in `eventlistview.cpp`) maps string and numeric keys to live datamodel objects. Unset optional fields throw `Core::ValueException` so null semantics work correctly with the leparser.

## Configuration

Rules are defined under `eventlist.highlight.*` in `global.cfg` so all tools share them. Parameter descriptions added to `global_gui.xml`.

Same condition syntax as documented in SeisComP/main#115.

### DIFF
--- a/libs/seiscomp/gui/datamodel/eventlistview.cpp
+++ b/libs/seiscomp/gui/datamodel/eventlistview.cpp
@@ -3334,32 +3334,26 @@ class EventKeyValueContext : public Utils::V2::LeKeyValueContext {
 
 		std::string getString(std::string_view key) const override {
 			if ( key == "type" ) {
-				try { return DataModel::EEventTypeNames::name(_event->type()); }
-				catch ( ... ) { throw Core::ValueException(); }
+				return DataModel::EEventTypeNames::name(_event->type());
 			}
 			if ( key == "typecertainty" ) {
-				try { return DataModel::EEventTypeCertaintyNames::name(_event->typeCertainty()); }
-				catch ( ... ) { throw Core::ValueException(); }
+				return DataModel::EEventTypeCertaintyNames::name(_event->typeCertainty());
 			}
 			if ( key == "evaluationstatus" || key == "status" ) {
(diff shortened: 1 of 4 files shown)


libs/seiscomp/gui/datamodel/eventlistview.cpp
```diff
@@ -3317,10 +3317,177 @@ EventListView::EventListView(Seiscomp::DataModel::DatabaseQuery* reader, bool wi
...
+		                     DataModel::Magnitude *magnitude)
+		    : _event(event), _origin(origin), _magnitude(magnitude) {}
+
+		std::string getString(std::string_view key) const override {
+			if ( key == "type" ) {
+				try { return DataModel::EEventTypeNames::name(_event->type()); }
+				catch ( ... ) { throw Core::ValueException(); }
+			}
+			if ( key == "typecertainty" ) {
+				try { return DataModel::EEventTypeCertaintyNames::name(_event->typeCertainty()); }
```
gempa-jabe: The cool thing is that you would not need try-catch as all datamodel method throw a `Core::ValueException` already ;)

gempa-jabe:
Pretty cool code-wise. I will merge it on Monday. You can clean up the code according to the above comment if you want otherwise we take it as-is and I will do it after more testing. I am happy with the current design, thanks a lot working on it.

comoglu:
The solution appears to be straightforward. I'll clean up the code now and spend the weekend with family. 👍

PR #197, opened 2026-05-08
Make Mwp amplitude and magnitude parameters configurable

comoglu:
## Summary

- **`amplitudes.Mwp.useFirstPeak`** (boolean, default `false`) — option to use the first local maximum of the doubly-integrated displacement after the P arrival, as specified in the original Tsuboi et al. (1995) algorithm, instead of the global maximum. Default `false` preserves backward compatibility. The global maximum can be inflated by later arrivals (PP, pP, surface reflections) for large events, contributing to Mwp saturation at M > 7.5.

- **`magnitudes.Mwp.estimateMw.a/b/stdError`** (double) — the Mw(Mwp) linear correction coefficients from Whitmore et al. (2002) (`a=1.186`, `b=-1.222`, `stdError=0.4`) were hardcoded with a `// Fixme` comment. They are now readable from `global.cfg`, allowing operators to apply updated calibrations from larger or more recent datasets without recompiling. Defaults are unchanged.

- **`global_Mwp.xml`** — new description file exposes `useFirstPeak` and `estimateMw.*` in scconfig.

## Backward compatibility

This PR is purely additive. Without any configuration changes the processor output is identical to the previous implementation in all cases.

## References

- Tsuboi et al. (1995, 1999) — Mwp algorithm
- Whitmore et al. (2002) — Mw(Mwp) correction, *Science of Tsunami Hazards* 20(4), 187–193

## Test plan

- [x] Build passes against SeisComP main source tree
- [ ] Build passes against SeisComP v7 source tree
- [x] `amplitudes.Mwp.useFirstPeak = true` confirmed via scamp debug log: `+ useFirstPeak = true`
- [ ] `global_Mwp.xml` parameters visible in scconfig

### DIFF
--- a/libs/seiscomp/processing/amplitudes/Mwp.cpp
+++ b/libs/seiscomp/processing/amplitudes/Mwp.cpp
@@ -20,7 +20,9 @@
 
 #include <seiscomp/processing/amplitudes/Mwp.h>
 #include <seiscomp/math/filter/iirintegrate.h>
-#include<seiscomp/math/filter/butterworth.h>
+#include <seiscomp/math/filter/butterworth.h>
+#include <seiscomp/logging/log.h>
+#include <seiscomp/config/config.h>
 
 #include <seiscomp/math/geo.h>
 
@@ -112,6 +114,26 @@ double Mwp_amplitude(int n, double *f, int i0, int *pos) {
 }
 
 
+// Tsuboi et al. (1995) original algorithm: first local peak after P onset,
+// not the global maximum.  Stops as soon as amplitude starts decreasing.
+double Mwp_first_peak_amplitude(int n, double *f, int i0, int *pos) {
+	double prev = 0.0;
+	*pos = i0;
+
+	for (int i = i0; i < n - 1; i++) {
+		double s = fabs(f[i]);
+		if (s >= prev) {
+			*pos = i;
+			prev = s;
+		}
+		else {
+			return prev;  // first local maximum found
+		}
+	}
+	return prev;
+}
+
+
 void Mwp_double_integration(int n, double *f, int i0, double fsamp) {
 	Mwp_integr(n, f, i0);
 	Mwp_integr(n, f, i0);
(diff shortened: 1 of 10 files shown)


libs/seiscomp/processing/amplitudes/Mwp.cpp
```diff
@@ -138,6 +160,29 @@ AmplitudeProcessor_Mwp::AmplitudeProcessor_Mwp()
...
+	if ( !AmplitudeProcessor::setup(settings) ) {
+		return false;
+	}
+
+	// Read from localConfiguration (global.cfg plain key or scconfig module.trunk. prefix).
+	// settings.getBool() uses a namespaced lookup that doesn't match bare global.cfg keys.
+	const Seiscomp::Config::Config *cfg = settings.localConfiguration;
+	if ( cfg ) {
+		if ( !cfg->getBool(_useFirstPeak, "amplitudes.Mwp.useFirstPeak") )
+			cfg->getBool(_useFirstPeak, "module.trunk.amplitudes.Mwp.useFirstPeak");
```
gempa-jabe: We would never read that kind of parameter. `module.trunk.[global|NET|NET.STA]` is used to read from bindings (still a hack for local configuration overrides), e.g., `settings.getBool("amplitudes.Mwp.useFirstPeak")`. Local configuration is read as you specified in your first line.

libs/seiscomp/processing/amplitudes/Mwp.cpp
```diff
@@ -209,7 +254,9 @@ bool AmplitudeProcessor_Mwp::computeAmplitude(const DoubleArray &data,
 	delete hp;
 
 	// Amplitude in nanometers
-	amplitude->value = 1.E9*Mwp_amplitude(si2, _processedData.typedData(), si1, &onset);
+	amplitude->value = _useFirstPeak
+	    ? 1.E9*Mwp_first_peak_amplitude(si2, _processedData.typedData(), si1, &onset)
```
gempa-jabe: While rewriting the code, please use whitespaces around arithmetic operators: `1E9 * Mwp_...`.

libs/seiscomp/processing/magnitudes/Mwp.cpp
```diff
@@ -50,10 +83,39 @@ MagnitudeProcessor_Mwp::MagnitudeProcessor_Mwp()
...
+		return false;
+	}
+
+	// Read from localConfiguration (global.cfg plain key or scconfig module.trunk. prefix).
+	// settings.getDouble() uses a namespaced lookup that doesn't match bare global.cfg keys.
+	const Seiscomp::Config::Config *cfg = settings.localConfiguration;
+	auto readDouble = [&](const std::string &key, double &val) {
+		if ( !cfg ) return;
+		if ( cfg->getDouble(val, key) ) return;
+		cfg->getDouble(val, "module.trunk." + key);
```
gempa-jabe: This alternative is not correct, see above.

comoglu:
I need to exclude one of the change which is not backward compatible. Coming up.

PR #198, opened 2026-05-18
Fixes memory leak in case of exception.

raciner:
readPTR allocates memory to decode content. If the object to be decoded is of the wrong type, an exception is thrown without the memory being freed. In case of a misconfigured module, this will lead to uncontrolled memory growth.

This fix addresses the issue by catching the exception, freeing the memory and throwing it again.


gempa-jabe:
Thank you for spotting and fixing that. Is there a way that I can reproduce it?

gempa-jabe:
@cla-bot check

PR #203, opened 2026-07-23
[scolv] make eventedit.scripts.column.* work

luca-s:
Currently the special values added by eventedit.scripts.column.* are never visualized on the event edit tab of scolv

### DIFF
--- a/libs/seiscomp/gui/datamodel/eventedit.cpp
+++ b/libs/seiscomp/gui/datamodel/eventedit.cpp
@@ -2753,7 +2753,7 @@ void EventEdit::insertOriginRow(Origin *org) {
 		}
 
 		PublicObjectEvaluator::Instance().prepend(this, org->publicID().c_str(),
-		                                          org->typeInfo().className(), scripts);
+		                                          org->typeInfo(), scripts);
 	}
 
 	_originTree->insertTopLevelItem(0, item);


luca-s:
Just to provide more context, the issue was in `PublicObjectEvaluator`, which was [failing to load origins from the database](https://github.com/SeisComP/common/blob/73658f169273909c35f5c7a275f86ba35f4f9757/libs/seiscomp/gui/datamodel/publicobjectevaluator.cpp#L432) because the required RTTI was missing from `job.classType`. Instead, [a string was passed](https://github.com/SeisComP/common/blob/73658f169273909c35f5c7a275f86ba35f4f9757/libs/seiscomp/gui/datamodel/eventedit.cpp#L2756). This PR fixes that issue.

Interestingly, if the user had the same scripts configured in the event list tab, the special columns in the event edit tab would still be populated for the currently selected origin (and only that row). This happened because the [EventEdit::evalResultAvailable](https://github.com/SeisComP/common/blob/73658f169273909c35f5c7a275f86ba35f4f9757/libs/seiscomp/gui/datamodel/eventedit.cpp#L4450) callback was triggered by the `PublicObjectEvaluator` instance used by the event list tab. In that code path, the correct RTTI information was passed, so the evaluation completed successfully.

luca-s:
@gempa-jabe I think this may have gone unnoticed, but it is a very straightforward merge. Would you also mind backporting it to the V7 branch? It's a small change, but not seeing the special column values is a bit annoying.

luca-s:
Thank you!

